PR #3, opened 2015-12-08
Update gcc requirement.

Ms2ger:
This library expects `get_compiler()` to return a `Tool` rather than a tuple, which is correct starting in 0.3.17.


alexcrichton:
Thanks!

PR #4, opened 2016-01-07
Build on mingw

FredrikNoren:
Mingw produces a slightly different environment than msys


FredrikNoren:
Doesn't work for some reason :/ Saw the comment on https://github.com/alexcrichton/libz-sys/commit/a362c307afe1509d07f58eee46b59b95063b13e6 but I guess that refers to https://msys2.github.io/?

PR #7, opened 2016-06-09
Bump cargo version

larsbergstrom:
r? @alexcrichton 

We need a bump of the cargo version to pick up the Android cross-build fix for Servo.

---
This change is [Reviewable](https://reviewable.io/reviews/alexcrichton/libz-sys/7)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "libz-sys"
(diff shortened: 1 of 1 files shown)


larsbergstrom:
Also, please cargo publish after this lands so that we can pick it up in Servo :-)

PR #12, opened 2016-12-24
Use metadeps to specify pkg-config dependencies declaratively

joshtriplett:
This makes it easier for distribution packaging tools to generate
appropriate package dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,5 +16,8 @@ Bindings to the system libz library (also known as zlib).
 libc = "0.2"
 
 [build-dependencies]
-pkg-config = "0.3"
+metadeps = "1"
 gcc = "0.3.17"
+
+[package.metadata.pkg-config]
+zlib = "1.2.3.4"
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,4 @@
-extern crate pkg_config;
+extern crate metadeps;
 extern crate gcc;
 
 use std::env;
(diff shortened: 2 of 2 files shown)


alexcrichton:
Thanks for the PRs! (here and across other repos)

I'm somewhat uneasy about adding these dependencies so deep in others' dependency graphs, though. Would it be possible to just add this metadata for now and eschew the `metadeps` dependency?

joshtriplett:
The metadata would help, but in many cases I'd be concerned that it'd get out of sync.  (Less an issue with something as stable as zlib, but still.)

I've tried to minimize the amount of additional dependencies introduced by metadeps; in particular, I eliminated the pile of dependencies that error-chain's backtrace pulls in by disabling default-features.  Switching to metadeps now only pulls in four crates: metadeps itself, toml, rustc-serialize, and error-chain.

And I just realized that the toml crate has a feature to disable rustc-serialize support as well, so I just uploaded metadeps 1.1.1 with that disabled.

So, switching to metadeps only involves three crates: metadeps itself, toml, and error-chain.  Does that seem sufficiently uninvasive to consider?

joshtriplett:
@alexcrichton The problem is that we're trying to build packages in a mostly automatic fashion for numerous crates, and a freeform comment doesn't help an automated packaging tool.

Would eliminating the error-chain dependency help at all (reducing it to just metadeps and toml, which seems as minimal as possible)?

If that still seems like too much, then yes, having the metadata would help even if not actually used by the `build.rs` script.

alexcrichton:
This is a pretty low-down-the-stack dependency, so I'm wary to pick up even more deps as a result. TOML nowadays depends on serde itself which I'd prefer to avoid. I'd be fine though just adding the annotation and keeping them in sync.

joshtriplett:
@alexcrichton metadeps doesn't actually depend on a version of toml that pulls in serde. I made an internals post seeking further metadeps feedback, and I'd welcome your input: https://internals.rust-lang.org/t/metadeps-and-dependencies/5349

PR #17, opened 2017-05-11
use static or dynamic build of zlib from vcpkg

mcgoo:
Use zlib from vcpkg. This one is not behind a feature flag but the vcpkg build helper has similar opt-out mechanisms to pkg-config.

This would be required in order to merge https://github.com/alexcrichton/curl-rust/pull/166.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,6 @@ libc = "0.2"
 [build-dependencies]
 pkg-config = "0.3"
 gcc = "0.3.17"
+
+[target.'cfg(target_env = "msvc")'.build-dependencies]
+vcpkg = "0.2"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,9 @@ environment:
     MSYS_BITS: 32
   - TARGET: x86_64-pc-windows-msvc
   - TARGET: i686-pc-windows-msvc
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks! I wonder, is it possible to make this a platform-specific build dependency?

Also I'm fine in general deferring most regression testing to the vcpkg crate itself, so mind adding just one extra builder for vcpkg?

alexcrichton:
Looks great to me, thanks! Would it be possible to avoid committing the `VCPKG_PANIC` though?

mcgoo:
I removed the VCPKG_PANIC.

I got a really nice email out of the blue from one of the Microsoft developers working on Vcpkg and he pointed out that the static Vcpkg builds are built against a static crt and so without +crt-static, the static builds are suspect. I have changed the build helper to select the static/dynamic variant based on whether target-feature=+crt-static is in effect. I wanted to draw your attention to that since this is a squashed commit and that's not precisely what you just ok'd :) If you don't think that is a sensible way to go about it, I'm open to other ideas.

Let me know if you want me to add you or anyone as an owner of the vcpkg crate/repo, or transfer it at any point.

alexcrichton:
Nah this looks great to me, thanks so much for this!

PR #18, opened 2017-06-10
nmake chokes on MAKEFLAGS for jobserver (unmergeable repro)

mcgoo:
```
NMAKE : fatal error U1065: invalid option '-'
Stop.
Command exited with code 101
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -185,6 +185,7 @@ fn build_msvc_zlib(target: &str) {
 
     let nmake = gcc::windows_registry::find(target, "nmake.exe");
     let mut nmake = nmake.unwrap_or(Command::new("nmake.exe"));
+    nmake.env_remove("MAKEFLAGS");
     run(nmake.current_dir(dst.join("build"))
              .arg("/nologo")
              .arg("/f")


mcgoo:
`MAKEFLAGS` is `--jobserver-fds=[...]`
`cargo --version` is `cargo 0.20.0-nightly (bbfe9b3a9 2017-06-05)`

mcgoo:
Seems to be working for you?! I must have I messed something up - here is it failing for me. I'll take a look and get back to you.

https://ci.appveyor.com/project/mcgoo/libz-sys/build/1.0.23

mcgoo:
Ok, I just went over it again and it's definitely a real issue. I can't understand why a build of 78d0856 fails for me and works for you - maybe the jobserver doesn't always kick in or something?

alexcrichton:
Thanks!

Arnavion:
@alexcrichton Is this a permanent fix or just there to immediately unblock nightly users?

If this is the permanent fix, it seems to imply that every build script that uses nmake will have to beware that cargo will pass MAKEFLAGS to it, and perhaps defensively guard against some future version of cargo passing in even more env vars that might interfere with whatever the build script runs. As I said in the linked cargo issue, I think it'll be better for cargo to pass its own env vars (eg `CARGO_JOBSERVER`) and let build scripts decide what to do with them.

alexcrichton:
I'm not sure, haven't decided yet.

I want this fix in libz-sys unconditionally, though. It's not the first time I've run into this.

I plan on continuing discussion on the related Cargo issue.

Arnavion:
Cool, thanks!

PR #19, opened 2017-06-27
Use gmake on DragonFly as well

mneumann:
Note: This applies to all other BSDs as well. Hard coding it in this way
should be generally avoided if possible.


alexcrichton:
Thanks!

PR #24, opened 2017-08-10
Detect when -lz doesn't work and compile zlib from source && rerun if env changed

cssivision:
fix https://github.com/alexcrichton/libz-sys/issues/21 and https://github.com/alexcrichton/libz-sys/issues/20.

add a detect for condition when -lz doesn't work and compile from source. but i found the code for platform determination is a little confused.


cssivision:
[try_compile](https://github.com/alexcrichton/gcc-rs/blob/master/src/lib.rs#L691) func in gcc-rs is better, but has not been released.

PR #26, opened 2017-10-14
Set MAKEFLAGS to CARGO_MAKEFLAGS for the make command

Eijebong:


alexcrichton:
Thanks! Perhaps this could be --

```rust
if let Some(s) = env::var_os("CARGO_MAKEFLAGS") {
    cmd.env("MAKEFLAGS", s);
}
```

Eijebong:
Sure, I did it this way because it was less typing :p

build.rs
```diff
@@ -91,6 +91,10 @@ fn build_zlib() {
 fn make() -> Command {
     let cmd = if cfg!(any(target_os = "freebsd", target_os = "dragonfly")) {"gmake"} else {"make"};
     let mut cmd = Command::new(cmd);
+    if let Some(makeflags) = env::var_os("CARGO_MAKEFLAGS") {
+        cmd.env("MAKEFLAGS", makeflags);
+    }
```
alexcrichton: Should this perhaps be combined with the `cfg!` below, as they look quite related? In that I'd imagine that the windows cfg would come first, and only if we're not on windows do we try to set `MAKEFLAGS`?

Eijebong: Well It adds an indentation and if `cfg!(windows)` is true, setting the `MAKEFLAGS` is a noop since it's removed just after. I can put it in an else statement if you want though

alexcrichton: Er what I mean is --

```rust
// comment
if cfg!(windows) {
    cmd.env_remove(...);
} else if let Some(..) = .. {
    cmd.env(..., ...);
}
```

Eijebong: Oh, right, it's better

PR #28, opened 2017-11-06
Replace library source with Git submodule

GabrielMajeri:
Instead of storing the whole `zlib` source code in the repo, this pull request replaces it with a Git submodule.

It uses the latest version of `zlib` from Mark Adler's [GitHub repo](https://github.com/madler/zlib), and can be more easily updated.


alexcrichton:
Nice! I think AppVeyor needs to be updated to checkout submodules though?

GabrielMajeri:
Yes, although the i686 builds seems to always fail in the last few pull requests. It looks like GNU doesn't compile `zlib` with `-Werror` on. I'll make another pull request for that too.

alexcrichton:
Yeah the 32-bit MinGW build broke awhile ago and I haven't gotten around to fixing it, but if you know how to fix it that would be awesome!

PR #30, opened 2017-11-06
Don't run pkg-config on FreeBSD and DragonFlyBSD

myfreeweb:
Avoid the same `-L/usr/lib` problem as on macOS.


alexcrichton:
Thanks!

PR #31, opened 2017-11-06
Fix MinGW build

GabrielMajeri:
This pull request attempts to fix 32-bit MinGW builds, which fail due to `-Werror` being enabled.


GabrielMajeri:
Can't fix at this crate's level, opened a relevant issue here: alexcrichton/ctest#24

PR #33, opened 2017-11-27
fix rerun-if-env-changed

cssivision:
fix https://github.com/alexcrichton/libz-sys/issues/21

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -20,6 +20,7 @@ macro_rules! t {
 
 fn main() {
     println!("cargo:rerun-if-env-changed=LIBZ_SYS_STATIC");
+    println!("cargo:rerun-if-changed=build.rs");
     let host = env::var("HOST").unwrap();
     let target = env::var("TARGET").unwrap();
 
--- a/build.rs
+++ b/build.rs
@@ -19,6 +19,7 @@ macro_rules! t {
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I think some `rerun-if-changed` should also be printed though, right? (for files depended on)

cssivision:
from http://doc.crates.io/build-script.html, `rerun-if-changed=PATH` 
Normally build scripts are re-run if any file inside the crate root changes, but this can be used to scope changes to just a small set of files. 

if the build script itself (or one of its dependencies) changes, then it's rebuilt and rerun unconditionally, so cargo:rerun-if-changed=build.rs is almost always redundant (unless you want to ignore changes in all other files except for build.rs).

alexcrichton:
Yeah but the default behavior is turned off if any other `rerun-if-*` keys are printed (such as this env var), so we'll have to recover the behavior by manually informing about input files. I think also that pkg-config now has an option to turn on the env vars by default?

cssivision:
you mean env use `rerun-if-env-changed` will turn off behavior  behavior of env `rerun-if-changed`?

alexcrichton:
Yeah unfortunately usage of either key turns of the default behavior :(

cssivision:
ok, i will fix this.

alexcrichton:
Ok! While this also depends on files in zlib itself I think printing out just the build script is fine for now.

PR #38, opened 2018-08-26
Set CHOST env variable when cross compiling

srijs:
Currently, cross-compiling this crate does not work out of the box on OSX, where `libz` is trying to use libtool to create the archive.

As a workaround, I've been setting the `CHOST` env variable in my build scripts, but that is less ideal IMO.

This PR automates setting `CHOST` when host and target triple differ. For reference, the option is documented here: https://github.com/madler/zlib/blob/master/configure#L33

EDIT: It looks like `CHOST` is not `zlib`-specific, so maybe this fix should land in the `cc` crate instead? Let me know which one you'd prefer!

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -55,29 +55,22 @@ fn main() {
                want_static) &&
               !target.contains("windows-gnu") &&
               !target.contains("android") {
-        build_zlib();
+        build_zlib(&host, &target);
     } else {
         println!("cargo:rustc-link-lib=z");
     }
 }
 
-fn build_zlib() {
+fn build_zlib(host: &str, target: &str) {
     let src = env::current_dir().unwrap().join("src/zlib");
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I've actually never heard of `CHOST` before... It seems like there's quite a number of "standards" for specifying cross compilations! 

Can you gist the error message that you were seeing before this PR as well?

srijs:
Heh, I hadn't either before I ran into this problem!

Here's the relevant part of the build log: https://gist.github.com/srijs/937da4c41e567f94810a367b112950ef

To me it looks like it's trying to use the OSX system `libtool` to create the static library, which doesn't work as it does not know how to deal with elf object files.

alexcrichton:
Geez what a bummer! In any case this looks good for this crate itself, thanks!

PR #45, opened 2018-10-12
Define _C99_SOURCE for iOS targets

### LINKED ISSUE #44: Regression building iOS targets between 1.0.20 --> 1.0.21+
Looking at the commits I'm guessing this has something to do with the rework to use `cc` crate for building libz, but for the life of me I can't figure out why.

Targets I'm building for (and results on 1.0.21, which are the same afaict to master branch):
- `aarch64-apple-ios` fails trying to resolve snprintf of all things [see gist output](https://gist.github.com/tiras-j/2bbc0dd1c57f7f54a657dcfa52244a81)
- `armv7-apple-ios` gives warnings but no [failure](https://gist.github.com/tiras-j/a9f56a8cf333820c4dfd85f122d0355a)
- `armv7s-apple-ios` has the same [warnings](https://gist.github.com/tiras-j/e222e7310e87ab8c85d3faf160f452f6)

What's crazy to me is I've looked at the libraries in those sysroot paths and libc.tbd absolutely contains `_snprintf` and the other symbols (I mean... no surprise there). Furthermore writing an incredibly simple c program that uses stdio + snprintf etc and compiling with the identical cross-compilation arguments (less the -D options since I'm not using them, but then again perhaps those are influencing the compilation somehow?) compiles just fine (at least for `aarch64-apple-ios`). 

Any thoughts on what could be going on here?

tiras-j:
It seems that snprintf and family are feature-gated behind either
`_POSIX_C_SOURCE=200112L`, `_C99_SOURCE`, or `__cplusplus` within the iOS 12
SDK stdio.h decls. This change gives successful cross-compilation
builds (from host `x86_64-apple-darwin`) for:
* `aarch64-apple-ios`
* `armv7-apple-ios`
* `armv7s-apple-ios`
* `x86_64-apple-ios`
* `i386-apple-ios`

Fixes #44

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -106,6 +106,9 @@ fn build_zlib(cfg: &mut cc::Build, target: &str) {
         cfg.define("_POSIX_SOURCE", None);
         cfg.flag("-fvisibility=hidden");
     }
+    if target.contains("ios") {
+        cfg.define("_C99_SOURCE", None);
+    }
     if target.contains("solaris") {
         cfg.define("_XOPEN_SOURCE", "700");
     }


tiras-j:
This has got me wondering about all the compilation dependencies between different standards declarations. I _think_ this was introduced by adding the `-D_POSIX_SOURCE` which defaults to defining `-D_POSIX_SOURCE=1` which itself implies `-D_POSIX_C_SOURCE=198808L` which maps directly to `__DARWIN_C_LEVEL` since that's simply `#define __DARWIN_C_LEVEL  _POSIX_C_SOURCE` and thus the `snprintf` family don't get included in stdio.h.

I'm a little worried how deep this rabbit-hole will go... Although looking closer at the cdef.h the solaris fix of setting `_XOPEN_SOURCE=700L` would solve the issue the same way, as it triggers undefining `_POSIX_C_SOURCE` which means `__DARWIN_C_LEVEL` defaults to >=200112L.

Unfortunately I don't know the implications of `_C99_SOURCE` vs `_XOPEN_SOURCE=700L`, but they both seem to work /shrug

PR #46, opened 2018-10-19
Add 'static' feature to build a statically linked crate

imp:
There is a way to build statically linked crate right now, however, it requires the environment variable `LIBZ_SYS_STATIC=1`. This can be suboptimal when as it makes the build to not be self-contained.
This PR adds feature `static` that triggers the same processing as the env var.


alexcrichton:
Thanks!

alexcrichton:
Sure thing, can do!

Issue #49, opened 2019-02-07
pkg_config always adds system-wide path to rustc-link-search


rinon:
Could we get `.print_system_libs(false)` added to the pkg_config `Config`? libz-sys is far down the dependency chain from my package and adds `/usr/lib` to the search path. This means that the linker grabs another library we want to link against from `/usr/lib` instead of our non-system search path that we add using `rustc-link-search`.

This is the same basic issue as https://github.com/sfackler/rust-openssl/issues/447 and https://github.com/alexcrichton/pkg-config-rs/issues/11

Thanks!

alexcrichton:
Thanks for the report! I don't think the answer is to not print out `-lz`, but to just not print out `-L /usr/...`. Currently I don't think there's an easy way to do that.

rinon:
If I understand `Config.print_system_libs` correctly, it just sets `PKG_CONFIG_ALLOW_SYSTEM_LIBS=1` for `pkg-config`, which just skips printing `-L /usr/lib` according to the man page:
> PKG_CONFIG_ALLOW_SYSTEM_LIBS
Don't strip -L/usr/lib out of libs

alexcrichton:
Ok seems like it just needs a PR then!

joshtriplett:
Merged.

PR #50, opened 2019-02-13
Don't let pkg-config add system lib dirs to the search path

rinon:
Adding system-wide library directories to the linker search path causes these
directories to be searched before other paths added by later crates. If a
library is present in the system-wide directory and a later crate wants to
specifically link against a different version of that library in another path,
the linker will choose the library from the first search directory it
finds. If the linker doesn't find a library in any of the specified search
directories, it falls back on system-wide paths, so we don't need to print the
path we found zlib in if it is in one of those system paths.

This resolves issue #49.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,9 +28,18 @@ fn main() {
        !target.contains("msvc") && // pkg-config just never works here
        !(host_and_target_contain("apple") ||
          host_and_target_contain("freebsd") ||
-         host_and_target_contain("dragonfly")) &&
-        pkg_config::Config::new().cargo_metadata(true).probe("zlib").is_ok() {
-        return
+         host_and_target_contain("dragonfly"))
+    {
+        // Don't print system lib dirs to cargo since this interferes with other
+        // packages adding non-system search paths to link against libraries
+        // that are also found in a system-wide lib dir.
+        let zlib = pkg_config::Config::new()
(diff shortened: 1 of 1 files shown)


alexcrichton:
@joshtriplett can you take a look at this?

joshtriplett:
@alexcrichton Why does pkg-config-rs default to allowing system library directories by default, rather than sticking with the default behavior of the pkg-config executable and stripping them out?

This seems like what *any* library should be doing, not just libz-sys.

rinon:
Agreed, I'd prefer that to be the default, but @alexcrichton didn't want to do this this because it changes the behavior of pkg-config-rs (https://github.com/rust-lang/pkg-config-rs/pull/35). There's a longer discussion in https://github.com/rust-lang/pkg-config-rs/issues/11.

alexcrichton:
@joshtriplett way long ago when it first came up I didn't want to maintain a list of "what are the system library directories" for all platforms, but times change!

rinon:
Why would `pkg-config-rs` need to maintain a list? pkg-config by default filters the list, do we need to duplicate that in Rust?

rinon:
Just wanted to ping this PR and see if it can be merged.

joshtriplett:
Looks reasonable; thank you for the patch.

PR #54, opened 2019-08-30
Enable shared building on musl

Cogitri:
I'm honestly unsure why that automagic in there is even needed, we have a `static` feature to set this, or if we can't find anything via pkg-config we'll do static linking anyway with the vendored version so this just limits the choice users have.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -58,7 +58,6 @@ fn main() {
         target.contains("pc-windows-gnu") ||
         want_static ||
         target != host ||
-        target.contains("musl")
     {
         return build_zlib(&mut cfg, &target);
     }


build.rs
```diff
@@ -58,7 +58,6 @@ fn main() {
         target.contains("pc-windows-gnu") ||
         want_static ||
         target != host ||
-        target.contains("musl")
```
joshtriplett, line 61: This doesn't appear to be build-tested; you've removed this line but not the `||` on the previous line.

Cogitri: Oh, I did test it but then submitted the patch via the webUI and apparently messed the editing up. Thanks for noticing.

PR #58, opened 2019-12-20
make static build compatible with cmake

BusyJay:
Common cmake script uses builtin `FindZLIB` module to find where zlib is
installed. The cmake crate allows `register_dep` to hint cmake to search
the right place. libz-sys puts the library under build directory, which
is however not the place where cmake is going to search. So this PR
corrects the behavior by moving the built library to lib directory.

Another option is to set the root directory to build instead, but the
layout may not look good.

It has been an issue for downstream projects like tikv/rust-rocksdb#303
and tikv/grpc-rs#419.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -91,9 +91,9 @@ fn main() {
 
 fn build_zlib(cfg: &mut cc::Build, target: &str) {
     let dst = PathBuf::from(env::var_os("OUT_DIR").unwrap());
-    let build = dst.join("build");
+    let lib = dst.join("lib");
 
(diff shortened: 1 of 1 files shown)


joshtriplett:
@BusyJay This seems potentially reasonable; however, it assumes the library will be named `libz.a`, which won't work with the Windows linker. Rather than renaming the library after it's built, could you instead change the destination directory to build it in the right place to begin with?

joshtriplett:
This looks much better, thank you! I'll look at merging this later this week.

joshtriplett:
"later this week", famous last words. Sorry for the *exceedingly* long delay.

PR #60, opened 2020-03-04
Allow building zlib without libc

RReverser:
Adds a `libc` feature which is enabled by default, but opting out from it allows to build zlib with Z_SOLO mode.

This mode removes dependency on any external libraries including libc at the cost of eliminating some high-level functions like gz*, compress* and uncompress, and requiring embedder to provide memory allocation routines to deflate and inflate.

This is useful for targets without libc support such as embedded targets or wasm32-unknown-unknown, as well as for embedding in crates that provide own allocation anyway.

An an example of the latter, when using this mode from a local branch of flate2, the output artifact of `cargo build --release --example compress_file --features zlib` in flate2 goes down by 7KB, even though all the functionality is still available as before.

Note: the reason we're introducing a feature `libc` and not a more natural variant `solo` is because Cargo takes features from all dependencies into the account. That way, if we had at least one dependency that uses libz-sys with `solo` feature, then all other dependencies would also get a Z_SOLO build, which might break their expectations if they depend on other high-level functions. Instead, by providing a `libc` feature, we guarantee that we build in Z_SOLO mode only if none of the dependencies needs the full zlib. For that, all such dependencies must use `default-features = false`.

This build also removes a dependency on the Rust libc crate, because otherwise we would need to disable it conditionally per-target, even though we only need types available in standard library under `std::os::raw` and using them allows to get rid of one more dependency altogether.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,6 @@ categories = ["external-ffi-bindings"]
 [workspace]
 members = ["systest"]
 
-[dependencies]
-libc = "0.2.43"
-
 [build-dependencies]
 pkg-config = "0.3.9"
 cc = "1.0.18"
@@ -27,9 +24,16 @@ cc = "1.0.18"
 vcpkg = "0.2"
 
 [features]
+default = ["libc"]
 # When building from source, enable building optimized assembly routines. As
 # noted in `contrib/README.contrib`, these routines are experimental and not
 # vetted, use at your own risk!
 asm = []
 # Enable this feature if you want to have a staticly linked libz
 static = []
+# When this feature is disabled, zlib will be built in Z_SOLO mode which
+# removes dependency on any external libraries like libc at the cost of
+# eliminating some high-level functions like gz*, compress* and
+# uncompress, and requiring embedder to provide memory allocation
+# routines to deflate and inflate.
+libc = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,14 @@ categories = ["external-ffi-bindings"]
 [workspace]
 members = ["systest"]
 
+[dependencies]
+# When this feature is disabled, zlib will be built in Z_SOLO mode which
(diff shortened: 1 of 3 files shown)


RReverser:
Looks like I missed one Unix-specific type (`off_t`) that does depend on libc. I guess will have to depend on libc conditionally after all :(

RReverser:
Should be fixed now.

alexcrichton:
Sorry I don't have the time to maintain this crate any more, I've handed it off to @joshtriplett

RReverser:
Oh, apologies for pinging then! I figured it was you from the commit history and ongoing (?) maintenance of flate2-rs.

@joshtriplett Could you please review this? Primary motivation is to get this crate working on pure Wasm target (`wasm32-unknown-unknown`), but, as mentioned in examples, it also benefits other targets and dependants like flate2-rs.

RReverser:
Sorry for the ping, but I'd really love some feedback on this to move forward with making PRs to other dependents.

RReverser:
@alexcrichton @joshtriplett Sorry, but it's been almost a month, and the change is fairly small... could someone please take a look and merge it if possible?

alexcrichton:
I have not resumed maintainership in the interim month, so sorry but I still can't help.

fennerm:
Also running into this issue in a wasm project. @joshtriplett sorry pinging you once more..

joshtriplett:
This looks reasonable to me. Sorry for the excessive delay in reviewing.

RReverser:
@joshtriplett Yay, thanks for merging! Could you publish a new release?

joshtriplett:
Followup: I ended up tweaking this a little, so that it doesn't expose functions or types that aren't defined when built without libc: https://github.com/rust-lang/libz-sys/commit/3dc5e8018063b9c053aaf12b986a469943bb3aab

PR #61, opened 2020-03-28
Format code using 'cargo fmt'

Atul9:


joshtriplett:
I ran `cargo fmt` on the current version and committed the result, to avoid conflicts.

Thanks for the suggestion!

PR #68, opened 2020-08-21
Use `ctest2` instead of the original to fix warnings

JohnTitor:
I saw Josh opened this issue: https://github.com/gnzlbg/ctest/issues/99
That's because we use ctest v0.1 here, it's outdated a bit. We could fix this issue by updating it to v0.2 but it will emit uninitialized errors, like this: https://github.com/JohnTitor/libz-sys/runs/1014401885
So, if you want to get rid of the warnings, I'd recommend using ctest2, which is a fork of ctest and has some fixes: https://github.com/JohnTitor/ctest2
I created this to use on the libc crate, but I hope this helps here as well :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,5 +58,5 @@ stock-zlib = []
 # Deprecated: the assembly routines are outdated, and either reduce performance
 # or cause segfaults.
 asm = []
-# Enable this feature if you want to have a staticly linked libz
+# Enable this feature if you want to have a statically linked libz
 static = []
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -58,5 +58,5 @@ stock-zlib = []
 # Deprecated: the assembly routines are outdated, and either reduce performance
 # or cause segfaults.
 asm = []
-# Enable this feature if you want to have a staticly linked libz
+# Enable this feature if you want to have a statically linked libz
```
joshtriplett, line 61: Would you mind splitting this into a separate commit?

joshtriplett:
One minor nit, and otherwise, I'd love to merge this. Thanks for the fix!

Cargo.toml
JohnTitor: Sure, added 2c3472526427a3763abb3675993500e819cdaefe.

joshtriplett:
Thank you!

PR #75, opened 2021-02-15
Haiku: always link to the system libz.so.

nielx:
When rustc and cargo are built for Haiku, they are cross-compiled from Linux.
The current logic of the build script decides to build from source when
cross-compiling. On Haiku, this is unnecessary. We can safely assume that
libz.so is on the target system. Therefore, like with Android, we can always
link to it.


joshtriplett:
Seems reasonable.

PR #78, opened 2021-04-02
Disable forced zlib vendoring on musl

Logarithmus:
I use Void Linux with `musl` as system libc (I don't use glibc at all). I'm trying to make a package for [wezterm](https://github.com/wez/wezterm). `wezterm` depends on `libz-sys`, and it fails to build because it tries to build `zlib` from source and it obviously doesn't find them.

So I suggest removing this hardcoded vendoring for musl, because your assumptions about musl being used for embedded statically build stuff is wrong. There are at least 3 distros from top of my mind which use musl: Void Linux, Alpine Linux, Adélie Linux.

Another option would be downloading `zlib` or `zlib-ng` sources tarball, unpacking it into the right location before starting to build `wezterm`. But it seems hacky and harder than removing single line from `libz-sys`'s `build.rs`.

See also: https://github.com/rust-lang/libz-sys/pull/54

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -70,7 +70,6 @@ fn main() {
         || target.contains("pc-windows-gnu")
         || want_static
         || target != host
-        || target.contains("musl")
     {
         return build_zlib(&mut cfg, &target);
     }


joshtriplett:
This does seem entirely reasonable.

I suspect that incorporating this change is likely to cause downstream breakage for many people, who have been using libz-sys without explicitly specifying the "static" feature, and have been relying on this particular line to force static linking when using musl.

Despite that, I do think that incorporating this change is the right thing to do. But I also think it's going to take some clear communication so that people currently doing static linking with musl will need to update their `Cargo.toml` to explicitly enable the `static` feature.

I'm going to merge this change, but before the next release, we'll need to communicate this widely.

joshtriplett:
I ended up going ahead and shipping this, because it will still statically link if you're cross-compiling. So, this change will only affect people who are running natively on a musl-based Rust toolchain, but who still want to statically link rather than using their system musl and zlib.

joshtriplett:
https://www.reddit.com/r/rust/comments/n43pcb/psa_libzsys_on_musl_no_longer_links_statically_by/
https://users.rust-lang.org/t/psa-libz-sys-on-musl-no-longer-links-statically-by-default-unless-cross-compiling/59300

PR #81, opened 2021-06-06
Darwin: don't build zlib if universal binaries are available

MarcusCalhoun-Lopez:
A 32-bit build on a 64-bit macOS system is considered cross compiling.
However, macOS supports "universal" binaries, so prebuilt binaries *are* installed.
The same is true of x86_64 build on a aarch64 macOS system.


joshtriplett:
https://github.com/rust-lang/libz-sys/pull/85 addresses this.

Issue #86, opened 2021-12-03
libz-sys is incompatible with Visual Studio 2022 (unsupported or unknown VisualStudio version: 17.0)


zulc22:
```
error: failed to run custom build command for `libz-sys v1.1.3`

Caused by:
  process didn't exit successfully: `C:\Users\kscot\OpenGMK\target\release\build\libz-sys-27377adaf726fbab\build-script-build` (exit code: 101)
  --- stdout
  cargo:rerun-if-env-changed=LIBZ_SYS_STATIC
  cargo:rerun-if-changed=build.rs

  --- stderr
  thread 'main' panicked at '

  unsupported or unknown VisualStudio version: 17.0
  if another version is installed consider running the appropriate vcvars script before building this crate
  ', C:\Users\kscot\.cargo\registry\src\github.com-1ecc6299db9ec823\cmake-0.1.45\src\lib.rs:797:25
  stack backtrace:
     0:     0x7ff6137010cf - <unknown>
     1:     0x7ff613719d8a - <unknown>
     2:     0x7ff6136fac78 - <unknown>
     3:     0x7ff613703a36 - <unknown>
     4:     0x7ff61370351c - <unknown>
     5:     0x7ff613704095 - <unknown>
     6:     0x7ff613625c80 - <unknown>
     7:     0x7ff61362ec4f - <unknown>
     8:     0x7ff613625b8c - <unknown>
     9:     0x7ff61362bbb6 - <unknown>
    10:     0x7ff61362825b - <unknown>
    11:     0x7ff6136227c8 - <unknown>
    12:     0x7ff613621198 - <unknown>
    13:     0x7ff613623db6 - <unknown>
    14:     0x7ff613623787 - <unknown>
    15:     0x7ff61362578c - <unknown>
    16:     0x7ff613700a17 - <unknown>
    17:     0x7ff613625770 - <unknown>
    18:     0x7ff613623296 - <unknown>
    19:     0x7ff61371e0e4 - <unknown>
    20:     0x7ffc458e54e0 - BaseThreadInitThunk
    21:     0x7ffc468a485b - RtlUserThreadStart```

joshtriplett:
With #87 merged, this should be fixed. Please feel free to open an issue if you encounter a bug with the new version.

PR #87, opened 2022-02-22
Use the latest build deps to fix builds on windows-2022

lespea:
Should fix https://github.com/rust-lang/libz-sys/issues/86

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,9 +25,9 @@ members = ["systest"]
 libc = { version = "0.2.43", optional = true }
 
 [build-dependencies]
-pkg-config = "0.3.9"
-cc = "1.0.18"
-cmake = { version = "0.1.44", optional = true }
+pkg-config = "0.3.24"
+cc = "1.0.73"
+cmake = { version = "0.1.48", optional = true }
 
 [target.'cfg(target_env = "msvc")'.build-dependencies]
 vcpkg = "0.2"


lespea:
Seems like it's trying to build x64 library on the 686 targets?  I'll have to dig into this later tonight I was hoping it was a simple fix.

```
D:\a\libz-sys\libz-sys\target\i686-pc-windows-msvc\debug\deps\liblibz_sys-b077f531860f8372.rlib : warning LNK4272: library machine type 'x64' conflicts with target machine type 'x86'
```

Byron:
It breaks for me _now_ as well, but that might be unrelated to what this PR is fixing as this issue surfaced apparently with libz-sys 1.1.4.

I get `Error: could not find native static library `zlib`, perhaps an -L flag is missing?`, [here is the full log](https://github.com/Byron/gitoxide/runs/5456731713?check_suite_focus=true#step:7:353). Sorry for the noise if it's unrelated entirely.

joshtriplett:
@lespea I'd appreciate your help fixing this issue. I think I've addressed all the other issues in CI other than this.

lespea:
@joshtriplett I just made a new PR based off of the latest main branch: https://github.com/rust-lang/libz-sys/pull/92

lespea:
@Byron you might need to install it via vcpkg?  Something like `vcpkg install zlib:x64-windows-static-md zlib:x86-windows-static` or `vcpkg install zlib:x64-windows-static-md zlib:x86-windows-static`.  I don't code in windows much but I had to install those when testing locally on my windows box.  I'm not sure which one exactly you need (`-md` or not) but obviously just choose the correct one for the architecture you're building.

joshtriplett:
@lespea Thanks for the updated PR! I've merged that one to fix Windows builds of zlib-ng.

Byron:
Thanks for the suggestions, @lespea , installing `zlib` like you suggested did indeed work for me :)!

PR #90, opened 2022-03-08
Install MinGW using msys2/setup-msys2

joshtriplett:


joshtriplett:
This seems to fix everything except the issue observed in https://github.com/rust-lang/libz-sys/pull/87 .

PR #92, opened 2022-03-11
Fix zlib-ng win32 builds

lespea:
I'm pretty sure this is the correct way to set this; it seems to pass CI anyway.  Not really an area I'm an expert in...  If there's a better way to do this lmk.


joshtriplett:
Thanks! Looks like this is passing CI.

PR #94, opened 2022-04-22
Fix static wasm build without wasi

### LINKED ISSUE #95: Doesn't compile for WebAssembly with `static` feature
See #94.

tbu-:
`libc` doesn't have `off_t` for WebAssembly without WASI, so it falls
back to `long`.  Do the same in this crate to fix the build error when
trying to statically build libz for WebAssembly.

Rough code from `zconf.h` declaring `z_off_t`:
```c++
#ifndef Z_SOLO
#  if defined(Z_HAVE_UNISTD_H) || defined(_LARGEFILE64_SOURCE)
// [...]
#    ifndef z_off_t
#      define z_off_t off_t
#    endif
#  endif
#endif

// [...]

#ifndef z_off_t
#  define z_off_t long
#endif
```

Fixes #95.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,8 +21,11 @@ pub type voidpf = *mut c_void;
 pub enum gzFile_s {}
 pub enum internal_state {}
 
-#[cfg(feature = "libc")]
+#[cfg(all(feature = "libc", not(all(target_family = "wasm", target_os = "unknown"))))]
 pub type z_off_t = libc::off_t;
+
+#[cfg(all(feature = "libc", all(target_family = "wasm", target_os = "unknown")))]
+pub type z_off_t = c_long;
 
 #[repr(C)]
 #[derive(Copy, Clone)]


joshtriplett:
@tbu- I'd prefer not to special-case WebAssembly here. Could you, instead, submit a PR to the libc crate defining off_t for WebAssembly targets?

tbu-:
@joshtriplett There's no `off_t` for WebAssembly, as far as I understand it. `off_t` is defined by POSIX, and WebAssembly doesn't implement POSIX. zlib defines `off_t` to be `long` when `off_t` isn't defined. It seems to me that the handling needs to be in zlib-sys.

joshtriplett:
@tbu- AFAICT WASI defines [`filesize`](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-filesize-u64) and [`filedelta`](https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-filedelta-s64), which provide the equivalents of `off_t`. So it's defined that offsets are always 64-bit.

tbu-:
WASI does indeed define `off_t`:

https://github.com/rust-lang/libc/blob/f6df53fd694f6fc903058c765efc10d77725b31b/src/wasi.rs#L23

This is only included when targeting WASI though, not on all of WebAssembly:

https://github.com/rust-lang/libc/blob/f6df53fd694f6fc903058c765efc10d77725b31b/src/lib.rs#L150-L156

But WASI is just a possible environment of WebAssembly, libz also works without WASI, and in this case `libz-sys` (**but not libz itself**) fails to build. libz falls back to C's `long` type if `off_t` isn't present. Since libz does that, libz-sys probably should as well?

joshtriplett:
@tbu- Then in that case, can you please modify this to only apply for wasm targets with OS "unknown", but not WASI targets?

tbu-:
The `target_env` of unknown targets is actually the empty string `""` and not `"unknown"`. :o

tbu-:
> @tbu- Then in that case, can you please modify this to only apply for wasm targets with OS "unknown", but not WASI targets?

Done. Thanks for your patience and working with me. :) The CI failure is unrelated, it's a 500 error from crates.io.

Should I add a CI job testing this? It works locally.

src/lib.rs
```diff
@@ -21,9 +21,12 @@ pub type voidpf = *mut c_void;
 pub enum gzFile_s {}
 pub enum internal_state {}
 
-#[cfg(feature = "libc")]
+#[cfg(all(feature = "libc", not(all(any(target_arch = "wasm32", target_arch = "wasm64"), target_env = ""))))]
```
joshtriplett, line 24:
```suggestion
#[cfg(all(feature = "libc", not(all(target_family = "wasm", target_os = "none"))))]
```

src/lib.rs
```diff
@@ -21,9 +21,12 @@ pub type voidpf = *mut c_void;
...
 pub type z_off_t = libc::off_t;
 
+#[cfg(all(feature = "libc", all(any(target_arch = "wasm32", target_arch = "wasm64"), target_env = "")))]
```
joshtriplett, line 27:
```suggestion
#[cfg(all(feature = "libc", not(all(target_family = "wasm", target_os = "none"))))]
```

tbu-:
Done.

By experimenting, I found that I need `target_os = "unknown"` instead of `target_os = "none"`.

joshtriplett:
> Done.
> 
> By experimenting, I found that I need `target_os = "unknown"` instead of `target_os = "none"`.

Right, good catch. Mixed it up with `x86_64-unonowm-none`. 🤦

PR #96, opened 2022-04-27
Update zlib-ng to fix segfault due to unaligned access

joshtriplett:
zlib-ng commit a39e323a4db80a57feecf2ae212c08070234050c ("Added memory alignment compensation functions for users who may be using custom allocators that don't align on the same boundary zlib-ng expects.") fixes the segfault.

Update zlib-ng to the latest version that passes flate2's testsuite.

Also, prevent the zlib-ng build process from touching the network, by not building the zlib-ng testsuite (which we don't invoke). This will improve build time as well.


Mark-Simulacrum (approved):
Lightly skimmed upstream update. No strong concerns.

PR #119, opened 2023-04-28
Improve documentation for zlib-ng, including development workflow

joshtriplett:


Byron (approved):
Thanks for the additional support to get started more quickly.

I will probably have a couple of questions later, but will see to putting their answers into the READMEs if appropriate.

PR #131, opened 2023-05-29
Fix `zlib_installed` not using the correct flags

BlackDex:
The `zlib_installed` function does not use the correct compiler flags when checking if there already is a libz library pre-build.

This causes libz-sys to build the internal libz version, which might be different from the pre-build libz version, and might cause issues with other libraries which also used the pre-build libz wether or not it's static or shared.

I came across this when using my custom toolchain build images to build static musl binaries. And this crate seemed to build the built-in version anyways. Which is an other issue, not addressed in this PR.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,6 @@
 use std::env;
 use std::fs;
 use std::path::PathBuf;
-use std::process::Command;
 
 fn main() {
     println!("cargo:rerun-if-env-changed=LIBZ_SYS_STATIC");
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot for the fix! Even though I have no experience with the build system used, but judging from [the documentation of `to_command()`](https://docs.rs/cc/1.0.79/cc/struct.Tool.html#method.to_command) it should be the correct way to go from compiler tool to invocable command.

PR #136, opened 2023-06-27
Allow using vcpkg for all Windows targets, and use find_package

micolous:
`vcpkg-rs` doesn't support `-pc-windows-gnu` (mingw) _yet_, but its integration tests will try to build `libz-sys`, which currently prevents using `vcpkg` for non-`msvc` targets.

This change should have no effect for `libz-sys` right now, but unblocks that work.

This also switches from `vcpkg::lib_names`/`probe` to `vcpkg::find_package`, which reliably gets the library names. Previously, this had hard coded assumptions using MSVC-style library naming (no `lib` prefix).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,8 +34,6 @@ libc = { version = "0.2.43", optional = true }
 pkg-config = "0.3.9"
 cc = "1.0.18"
 cmake = { version = "0.1.44", optional = true }
-
-[target.'cfg(target_env = "msvc")'.build-dependencies]
 vcpkg = "0.2"
 
 [features]
--- a/build.rs
+++ b/build.rs
@@ -162,8 +162,7 @@ fn try_vcpkg() -> bool {
     // see if there is a vcpkg tree with zlib installed
     match vcpkg::Config::new()
         .emit_includes(true)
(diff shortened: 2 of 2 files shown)


Byron (approved):
Thanks a lot for tackling this!

To me this looks like an improvement all around as it reduces special cases that, at least according to CI, aren't needed (anymore).

I'd merge this, but would like another reviewer for good measure.

Byron:
I have rebased this PR onto `main` which contains additional tests against `flate2` which gives me all the confidence I need to merge this PR once CI is green.

micolous:
Thanks!

One addendum to the release notes for v1.1.10: this won't actually build for non-MSVC targets until https://github.com/mcgoo/vcpkg-rs/pull/52 is merged, and there's a new release of `vcpkg-rs` with it.

This PR just lays some of the ground work, and a release means I won't have to pin `libz-sys` in `vcpkg-rs`' CI any more 🎉

Byron:
Thanks for the clarification! I added the amendment [to the release notes](https://github.com/rust-lang/libz-sys/releases/tag/1.1.10), and am looking forward to seeing the `vcpkg-rs` improvements merged. Thanks for driving this!

PR #133, opened 2023-06-10
Update zlib-ng to 2.1.3

JakubOnderka:
zlib-ng 2.1.2 is the first stable release of the 2.1.x branch.

Most important optimizations and Enhancements:

* Decompression is a lot faster (56% faster measured on AVX2-capable x86-64)
* Compresson is improved for Level 9, at the cost of a little performance.
* Compression is improved for Level 3, by switching from deflate_fast to deflate_medium.

[Full changelog](https://github.com/zlib-ng/zlib-ng/releases/tag/2.1.2)

PR #139, opened 2023-07-23
increase CI confidence

Byron:
Test `flate2` against the locally available `libz-sys` code as well.


Byron:
It turns out that `zlib-ng` or `zlib-ng-compat` fail with the what's on `main`, with memory corruption, but what's in #133 seems to work thus far. Thus I will close this PR and apply the patch with the additional test script there before merging.

PR #141, opened 2023-07-24
Fix build on x86_64

JakubOnderka:
On same platform, library file is compiled into `lib64` directory. Rust compiler expects that library will be in `lib` directory. This fixes this by searing library both in `lib` and `lib64` directory.

### DIFF
--- a/build_zng.rs
+++ b/build_zng.rs
@@ -22,9 +22,14 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
 
     let includedir = install_dir.join("include");
     let libdir = install_dir.join("lib");
+    let libdir64 = install_dir.join("lib64");
     println!(
         "cargo:rustc-link-search=native={}",
         libdir.to_str().unwrap()
+    );
+    println!(
+        "cargo:rustc-link-search=native={}",
+        libdir64.to_str().unwrap()
     );
     let mut debug_suffix = "";
     let libname = if target.contains("windows") && target.contains("msvc") {


Byron:
Thanks for the fix!

Could you explain more about the origin of the issue that is fixed here? I'd like to understand if that was an issue that was recently introduced, or if it is a zero-day that now just happens to surface. Maybe it's a different platform that's causing this that we might want to try to add to CI?

Thanks for elaborating.

JakubOnderka:
I am actually not sure, but there was changed in cmake scripts in zlib-ng to 2.1. And the default cmake detection is a mess :D 
So I think if we want to test it in CI, we have to also test against different distribution than Ubuntu or Debian.

https://github.com/Kitware/CMake/blob/6194193a8a0a476539590aabbb27094d1aff136a/Modules/GNUInstallDirs.cmake#L222-L226

Byron:
Thanks a lot!

As CI is working, this means that having multiple search paths doesn't confuse `rustc`, so nothing should be able to break with this change. I also added the context you kindly provided to the merge commit for good measure.

Byron:
Right! Even though the breakage doesn't reproduce yet, it does exist and was introduced with the CMake upgrades in the more recent version of `zlib-ng`.

Anyway, [here is the new release](https://github.com/rust-lang/libz-sys/releases/tag/1.1.11).

Unfortunately it doesn't seem super trivial to change the linux distribution with `cross`, but maybe @JakubOnderka or @torokati44 can provide information on how to get such a previously failing distribution to run under GitHub actions CI. This way, regression tests could be added for this, increasing coverage.

torokati44:
Actually it was a fellow contributor who had an error related to `zlib.h` when running clippy on Windows, after updating to the (previous) latest release; and linked to this issue. I just butted in here with my comment above to hopefully help things move along for him. 😶
It seems his issue is still not fixed, he might report it soon.

Byron:
It's my first time to maintain a crate with 30m total downloads, and with the amount of users and platforms it's close to impossible to cover all bases. Thus, when PRs appear that fix something, I will try harder to be sure everything there is to know about it is disclosed before taking an action. Particularly, CI must be able to reproduce an issue and I don't think it's wise to proceed like I did thus far.

Issue #143, opened 2023-07-25
Build failure in msys2 in windows after updating from 1.1.9 to 1.1.10


Dinnerbone:
I thought (hoped) #141 would fix it but seemingly not.

When running `cargo clippy` on a project that indirectly depends on libz-sys, I can no longer build after libz-sys got updated to 1.1.10 (including 1.1.11). I get the following error:

```
error: failed to run custom build command for `curl-sys v0.4.64+curl-8.2.0`

Caused by:
  process didn't exit successfully: `C:\Users\Dinnerbone\workspace\personal\ruffle\target\debug\build\curl-sys-98d10b0290d5e12c\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-changed=curl
  cargo:root=C:\Users\Dinnerbone\workspace\personal\ruffle\target\debug\build\curl-sys-7017c58811c3cc06\out
  cargo:include=C:\Users\Dinnerbone\workspace\personal\ruffle\target\debug\build\curl-sys-7017c58811c3cc06\out\include
  cargo:static=1
  cargo:rustc-cfg=libcurl_vendored
  cargo:rustc-cfg=link_libnghttp2
  cargo:rustc-cfg=link_libz
  TARGET = Some("x86_64-pc-windows-msvc")
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-pc-windows-msvc")
  cargo:rerun-if-env-changed=CC_x86_64-pc-windows-msvc
  CC_x86_64-pc-windows-msvc = None
  cargo:rerun-if-env-changed=CC_x86_64_pc_windows_msvc
  CC_x86_64_pc_windows_msvc = None
  cargo:rerun-if-env-changed=HOST_CC
  HOST_CC = None
  cargo:rerun-if-env-changed=CC
  CC = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64-pc-windows-msvc
  CFLAGS_x86_64-pc-windows-msvc = None
  cargo:rerun-if-env-changed=CFLAGS_x86_64_pc_windows_msvc
  CFLAGS_x86_64_pc_windows_msvc = None
  cargo:rerun-if-env-changed=HOST_CFLAGS
  HOST_CFLAGS = None
  cargo:rerun-if-env-changed=CFLAGS
  CFLAGS = None
  cargo:rerun-if-env-changed=CRATE_CC_NO_DEFAULTS
  CRATE_CC_NO_DEFAULTS = None
[...]

PR #144, opened 2023-07-25
add cargo clippy runs to see if it can build successfully (#143)

Byron:
This hopefully reproduces #143.


Byron:
It seems to be hard to determine if clippy is available or not. But besides that, it's clear that this test doesn't reproduce.

PR #146, opened 2023-07-26
Revert "Run pkg-config on MSVC"

Byron:
This reverts commit 491bd1e137d63999e7a38370b6c0c05954b1079f.
Fixes https://github.com/rust-lang/libz-sys/issues/143.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libz-sys"
-version = "1.1.11"
+version = "1.1.10"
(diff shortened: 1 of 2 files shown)


nirbheek:
Copy-pasting my comment from the other issue:

Which zlib.pc is getting picked up while running on MSYS2? I think we need to investigate that and fix it instead of just disabling pkg-config again on windows.

Byron:
Closing as this fix was merged with a minor edit (so GitHub won't pick it up automatically).

Those who wish to follow up on @nirbheek copy-paste comment, here is the original: https://github.com/rust-lang/libz-sys/issues/143#issuecomment-1651178307 .

PR #157, opened 2023-09-12
Write more variables into zlib.pc

sagudev:
Added version, that is parsed from zlib.h and `includedir` and `libdir`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -137,11 +137,20 @@ fn build_zlib(cfg: &mut cc::Build, target: &str) {
     fs::copy("src/zlib/zconf.h", dst.join("include/zconf.h")).unwrap();
 
     fs::create_dir_all(lib.join("pkgconfig")).unwrap();
+    let zlib_h = fs::read_to_string(dst.join("include/zlib.h")).unwrap();
+    let version = zlib_h
+        .lines()
+        .find(|l| l.contains("ZLIB_VERSION"))
+        .unwrap()
+        .split("\"")
+        .nth(1)
+        .unwrap();
     fs::write(
         lib.join("pkgconfig/zlib.pc"),
         fs::read_to_string("src/zlib/zlib.pc.in")
             .unwrap()
-            .replace("@prefix@", dst.to_str().unwrap()),
+            .replace("@prefix@", dst.to_str().unwrap())
+            .replace("@VERSION@", version),
     )
     .unwrap();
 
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
Thanks for contributing.

Could you explain what the purpose of this change is? What's the benefit, what does it fix?
Thanks

sagudev:
Some C/C++ programs have configure scripts that check zlib versions. This PR just fills some of those placeholder so `zlib.pc` can be used by configure script.

build.rs
```diff
@@ -137,11 +137,22 @@ fn build_zlib(cfg: &mut cc::Build, target: &str) {
...
+        .split("\"")
+        .nth(1)
+        .unwrap();
     fs::write(
         lib.join("pkgconfig/zlib.pc"),
         fs::read_to_string("src/zlib/zlib.pc.in")
             .unwrap()
-            .replace("@prefix@", dst.to_str().unwrap()),
+            .replace("@prefix@", dst.to_str().unwrap())
+            .replace("@includedir@", "${prefix}/include")
```
Byron, line 153: Let's remove `includedir` and `libdir`, as these are generated and can as well be generated by the build system interpreting these files.

Byron (changes requested):
Thank you, I think more information shouldn't hurt.

I thought about this and have to consider it a backwards-incompatible change as this will crash hard if a header file provided by any system does not contain the desired information, information that previously wasn't provided.

To be save, please make the version optional. Thank you.

sagudev:
> I thought about this and have to consider it a backwards-incompatible change as this will crash hard if a header file provided by any system does not contain the desired information, information that previously wasn't provided.

Making zlib.pc is only done when compiling from source, which is part of this repo, so zlib.h will always contain that information, that's how zlib's build system is doing it. But better safe than sorry, so there will be no panic on version fail.

Byron (changes requested):
I am sorry for the earlier review - it was based on a browser-only session with a tiny context window. This made me miss that indeed, this change only affects the vendored version, but also that comes with `zlib.pc.in` which has `libdir` and `includedir` variables ready for substitution.

With that knowledge, I think the change is good the way it was *before* 😅, assuming that you are sure that the generated `includedir` and `libdir` variables will be correct on any system.

Thanks for bearing with me.

sagudev:
No worries, git stores previous commits too.

sagudev:
> assuming that you are sure that the generated includedir and libdir variables will be correct on any system.

If prefix is valid than includedir and libdir should be valid too.

Byron:
Thanks again!

PR #162, opened 2023-12-17
Update zlib and zlib-ng library

JakubOnderka:

PR #168, opened 2024-01-08
bump patch

### LINKED ISSUE #166: bump patch for crate
just a suggestion, as in msys2 we want to have fix for `*-gnullvm` target applied at stable so we can update it for needed packages (e.g. starship)

ognevnydemon:
so #162 will be applied
fixes #166


Byron:
Thanks a lot, a new release should be out shortly.

PR #174, opened 2024-01-17
Support `cc` as an optional alternative to `cmake`

Jake-Shadle:
While compiling [another project](https://github.com/Byron/gitoxide) I noticed I was not able to cross compile to x86_64-pc-windows-msvc, due to cmake. We've banned cmake in our projects both because it (generally speaking) complicates/makes impossible cross compilation, as well as just adding another dependency/point of failure/hindrance to contributions.

This is mainly a straight forward port of the cmake script(s) to cc but with one important difference, namely that target features are now respected.

With cmake, zlib-ng is probing the compiler/target headers for intrinsics/features/definitions, and importantly, can enable features irrespective of the target/features supplied to cargo/rustc. For example, `cargo build --features zlib-ng --target x86_64-unknown-linux-gnu` on my current machine will enable `avx2`, `sse2`, `sse3`, `sse4.2`, `pclmulqdq`, `xsave`, `avx512`, and `vpclmulqdq`, despite the only target features enabled by default for that target being `fxsr`, `sse`, and  `sse2`.

With the change to cc, I made it so that the target features selected by the user are what are actually respected, instead of generating binaries that won't work on CPUs without 1 or more features that the machine the binary was built on did have.

The one caveat to this is that avx512 target features are not currently stable in rustc, so I've added an additional feature that needs to be enabled for those features to be probed for, but maybe there is a more elegant way to support that in a way that makes sense to the end user.

I also didn't port the powerpc, s390, or riscv related optimization features since this PR is already quite large, but it should be trivial to add support in subsequent PRs.


Byron:
Thanks so much for tackling this! I agree that *less* `cmake` will be better for everyone.

Previously it was my understanding that `zlib-ng` ships with `cmake` as build system, and superseding it with an own implementation also means it needs to be maintained here.

For that reason I definitely won't be merging this PR without explicit support from the libs team, and I think it would be easiest to use [their channel](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs) to probe for additional opinions.

*(I also converted the PR into a draft while CI is failing)*

NobodyXu:
> With cmake, zlib-ng is probing the compiler/target headers for intrinsics/features/definitions, and importantly, can enable features irrespective of the target/features supplied to cargo/rustc. For example, cargo build --features zlib-ng --target x86_64-unknown-linux-gnu on my current machine will enable avx2, sse2, sse3, sse4.2, pclmulqdq, xsave, avx512, and vpclmulqdq, despite the only target features enabled by default for that target being fxsr, sse, and  sse2.

I see, that's probably why after upgrading libz-ng to latest release break our cargo-binstall cross compilation on aarch64 linux.

We use cargo-zigbuild for cross compilation and it was working until libz-ng is upgraded https://github.com/rust-cross/cargo-zigbuild/issues/210

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+    cfg.files(
+        files
+            .into_iter()
+            .map(|fname| format!("src/zlib-ng/{root}{fname}.c")),
+    );
+}
+
+/// Replicate the behavior of cmake/make/configure of stripping out the
+/// @ZLIB_SYMBOL_PREFIX@ since we don't want or need it
+fn strip_symbol_prefix(input: &Path, output: &Path, mut on_line: impl FnMut(&str)) {
```
NobodyXu, line 20: I suggest using trait object for faster compilation:

```suggestion
fn strip_symbol_prefix(input: &Path, output: &Path, on_line: & dyn mut FnMut(&str)) {
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+    let contents = fs::read_to_string(input).unwrap();
+    let mut h = fs::File::create(output).expect("failed to create zlib include");
```
NobodyXu, line 22: IMO it makes sense to use a buffer for the writes here

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+    root: Option<&'ctx str>,
+    enabled: Option<&'ctx BTreeSet<&'ctx str>>,
+    msvc: bool,
+    append: AppendState,
+    cache: Option<BTreeMap<u64, bool>>,
+}
+
+impl<'ctx> Ctx<'ctx> {
+    fn enable(&mut self, tf: TargetFeature) -> bool {
+        if let Some(enabled) = self.enabled {
```
NobodyXu, line 74: You can use a match here to avoid having two ifs:

```suggestion
        match self.enabled {
            Some(enabled) if !enabled.contains(tf.check) => false,
            _ => {
                self.push(tf);
                true
            }
        }
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+        };
+
+        cmd.arg(path);
+
+        let output = cmd.output().expect("failed to run compiler");
+
+        let result = if !output.status.success() {
+            println!("cargo:warning=failed to compile check '{tf:?}'");
+
+            for line in String::from_utf8(output.stderr).unwrap().lines() {
```
NobodyXu, line 139: I think panicking on non utf-8 character isn't very robust:

```suggestion
            for line in String::from_utf8_lossy(output.stderr).lines() {
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+    fn compile_check_root() -> PathBuf {
+        let mut pb = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+        pb.push("compile_check");
+        pb
+    }
+}
+
+impl<'ctx> Drop for Ctx<'ctx> {
+    fn drop(&mut self) {
+        let app = std::mem::replace(&mut self.append, Default::default());
```
NobodyXu, line 215:
```suggestion
        let app = std::mem::take(&mut self.append);
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+        fs::read_to_string("src/zlib-ng/zlib.pc.in")
+            .unwrap()
+            .replace("@prefix@", dst.to_str().unwrap())
+            .replace("@includedir@", "${prefix}/include")
+            .replace("@libdir@", "${prefix}/lib")
+            .replace("@VERSION@", &version),
+    )
+    .unwrap();
+
+    let dst = dst.to_str().unwrap();
```
NobodyXu, line 600: `dst` could contain non utf-i characters:

```suggestion
    let dst = dst.display();
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+    println!("cargo:root={dst}");
+    println!("cargo:rustc-link-search=native={}", lib.to_str().unwrap());
```
NobodyXu, line 602:
```suggestion
    println!("cargo:rustc-link-search=native={}", lib.display());
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
-use std::env;
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    env, fs,
+    hash::{Hash as _, Hasher as _},
+    io::Write as _,
+    path::{Path, PathBuf},
+};
+
+fn append(cfg: &mut cc::Build, root: Option<&str>, files: impl IntoIterator<Item = &'static str>) {
```
NobodyXu, line 9: Recommends to use trait object to speedup compilation:

```suggestion
fn append(cfg: &mut cc::Build, root: Option<&str>, files: &dyn mut Iterator<Item = &'static str>) {
```

build_zng.rs
```diff
@@ -1,58 +1,691 @@
...
+    files: &'static [&'static str],
+}
+
+struct Ctx<'ctx> {
+    cfg: &'ctx mut cc::Build,
+    root: Option<&'ctx str>,
+    enabled: Option<&'ctx BTreeSet<&'ctx str>>,
+    msvc: bool,
+    append: AppendState,
+    cache: Option<BTreeMap<u64, bool>>,
```
NobodyXu, line 69: Why not use:

```suggestion
    cache: Option<BTreeMap<TargetFeature, bool>>,
```

Why do you manually calculate the hash instead of using hashmap?

Hash can collide and without storing `TargetFeature`, you can't perform an equality check to make sure it isn't a , which HashMap already does the job for you.

I also find no reason to use HashMap over BTreeMap since 8MO there aren't enough elements for HashMap to outperform BTreeMap, it only slows down compilation since HashMap is more complex than BTreeMap and you would have to `derive(Hash)`

NobodyXu (changes requested):

build_zng.rs
Jake-Shadle: I don't understand the complaint here, you are both saying to use hashmap and not use hashmap?

build_zng.rs
Jake-Shadle: This is a copy of the same code in build.rs.

build_zng.rs
Jake-Shadle: This is a copy of what build.rs does.

build_zng.rs
Jake-Shadle: There is no case where a compiler would output non-utf8 strings, but sure.

build_zng.rs
Jake-Shadle: Any speedup would not be noticeable, I don't see the point.

joshtriplett:
> With cmake, zlib-ng is probing the compiler/target headers for intrinsics/features/definitions, and importantly, can enable features irrespective of the target/features supplied to cargo/rustc. For example, `cargo build --features zlib-ng --target x86_64-unknown-linux-gnu` on my current machine will enable `avx2`, `sse2`, `sse3`, `sse4.2`, `pclmulqdq`, `xsave`, `avx512`, and `vpclmulqdq`, despite the only target features enabled by default for that target being `fxsr`, `sse`, and `sse2`.

zlib-ng *should* only be doing that if given `WITH_NATIVE_INSTRUCTIONS`; otherwise, it should be using runtime detection, which uses these intrinsics but only after checking the CPU capabilities at runtime. If it's using features from the host CPU in any other circumstance, that's absolutely a bug.

build_zng.rs
NobodyXu: I'm saying that using hasher directly with a BTreeMap is wrong and you should use HashMap instead of hasher + BTreeMap.

However given the amount of data I don't think HashMap is necessary and BTreeMap might be faster and has less compilation overhead.

Jake-Shadle:
> zlib-ng should only be doing that if given WITH_NATIVE_INSTRUCTIONS; otherwise, it should be using runtime detection, which uses these intrinsics but only after checking the CPU capabilities at runtime. If it's using features from the host CPU in any other circumstance, that's absolutely a bug.

Yes, I was wrong about this, the probing for intrinsics just adds the source files and appropriate compiler flags if the compiler supports them, the runtime detection will only use them if the host CPU supports them.

I think if this PR would otherwise be accepted I can change the code to just do what zlib-ng does and enable all the source files/flags supported by the C compiler rather than relying on the target features to be enabled in rustc, but I'm not going to spend more time on this PR unless it's indicated it would be merged barring that change.

Jake-Shadle:
Is there any chance of getting a review on this? I'd rather close this than have it sitting.

NobodyXu:
@Jake-Shadle I'd recommend you to open a new topic in zulip channel https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates and cc those who works in libz-ng

wcampbell0x2a:
This would be amazing, as I would love to use this library for https://github.com/wcampbell0x2a/backhand/pulls, but don't want cmake in my build system.

joshtriplett:
I would love to see this happen, but also, this seems like a *massive* increase in complexity and maintenance burden for this crate, to duplicate the build system of upstream.

Is there any chance we could get upstream zlib-ng to add support for a non-cmake build system and then use that, so that we don't have to effectively rewrite their build system downstream and then maintain that rewrite as upstream's build system continues to change in the future?

Jake-Shadle:
Other than the `*.in` file transformations, all of the complexity in the build script beyond what a normal project has is simply due to the compiler probing for features and headers. IMO this is incredibly over conservative in an attempt to support really old compilers or "esoteric" targets that are already supported by vanilla zlib.

The easiest way to reduce complexity would just be to remove all of the probing and just....turn on everything for the target. If the user is using an extremely old compiler or targeting something that doesn't support all of the features, they can either 1. special case the build script to remove the specific feature(s) not supported by their concrete use-case 2. disable the zlib-ng feature to just use vanilla zlib instead. The build script could also just detect the failure and fallback to zlib automatically and print a warning about it, but generally I find a hard error to let the user know something went wrong is better so that they can either fix it or workaround it.

> Is there any chance we could get upstream zlib-ng to add support for a non-cmake build system and then use that, so that we don't have to effectively rewrite their build system downstream and then maintain that rewrite as upstream's build system continues to change in the future?
What would that be though? I don't see what reason they would have to support another one for a single downstream project.

NobodyXu:
> If the user is using an extremely old compiler or targeting something that doesn't support all of the features

IMO adding environment variable for these features make more sense

joshtriplett:
> all of the complexity in the build script beyond what a normal project has is simply due to the compiler probing for features and headers

That's not the only complexity or maintenance burden here. There's also the list of files (needs updating every time zlib-ng adds a new source file), and the list of symbols to define (needs updating every time zlib-ng's build system adds a new such symbol).

I'm not saying we shouldn't *consider* doing this. I'm observing that if we do, we change the maintenance burden from "upgrade to new upstream version of zlib-ng and make sure tests pass" to "upgrade to new upstream version of zlib-ng, audit the build system to see if they've added any new symbols, and port any logic needed to set those symbols".

> The easiest way to reduce complexity would just be to remove all of the probing and just....turn on everything for the target.

Defaulting to building zlib-ng in "probe for everything you know about at runtime" mode does seem like the right answer, particularly if it simplifies the build logic; if people want to hard-enable or hard-disable any features, or probe at compile time, perhaps we could control the use of cmake versus the non-cmake simplified build system via feature flags.

> I don't see what reason they would have to support another one for a single downstream project.

The same reason *this* project would consider doing so: to try to help their users who don't want to add a cmake dependency, at the expense of taking on additional complexity. The difference is that zlib-ng is better positioned to manage that complexity than we are.

I know at least one project that's considering a build system rewrite away from cmake for the exact same reason: aws-lc.

Jake-Shadle:
> That's not the only complexity or maintenance burden here. There's also the list of files (needs updating every time zlib-ng adds a new source file), and the list of symbols to define (needs updating every time zlib-ng's build system adds a new such symbol).

> I'm not saying we shouldn't consider doing this. I'm observing that if we do, we change the maintenance burden from "upgrade to new upstream version of zlib-ng and make sure tests pass" to "upgrade to new upstream version of zlib-ng, audit the build system to see if they've added any new symbols, and port any logic needed to set those symbols".

That's the normal amount of burden that any cc-based build script needs to do, looking at zlib-ng history there does seem to be more churn wrt build system affecting file/etc changes compared to the typical C project, but not to a degree that would make version upgrades take a significant longer period of time barring large version updates or a too long period of not updating.

> Defaulting to building zlib-ng in "probe for everything you know about at runtime" mode does seem like the right answer, particularly if it simplifies the build logic; if people want to hard-enable or hard-disable any features, or probe at compile time, perhaps we could control the use of cmake versus the non-cmake simplified build system via feature flags.

I think that's a reasonable compromise, that way if the simpler cc-based build fails the error can tell them to turn on the `cmake` feature to do the probing to support their limited environment/target but most won't need to do that.

> The same reason this project would consider doing so: to try to help their users who don't want to add a cmake dependency, at the expense of taking on additional complexity. The difference is that zlib-ng is better positioned to manage that complexity than we are.

They support regular make, presumably for those who don't want to use cmake, I just am not aware of any other build systems in the C space that don't add external dependencies (python, lua, etc) or are platform specific (bash scripts, vcxproj etc) that one would have any hope of proposing to them.

Byron:
> > Defaulting to building zlib-ng in "probe for everything you know about at runtime" mode does seem like the right answer, particularly if it simplifies the build logic; if people want to hard-enable or hard-disable any features, or probe at compile time, perhaps we could control the use of cmake versus the non-cmake simplified build system via feature flags.
> 
> I think that's a reasonable compromise, that way if the simpler cc-based build fails the error can tell them to turn on the `cmake` feature to do the probing to support their limited environment/target but most won't need to do that.

The above part of the conversation gave me an idea I want to share, right after sharing my disposition about this: I feel strongly about *not wanting* to chase `zlib-ng` build system changes to the point where I wont hit the merge button for that reason alone. However, I also wouldn't oppose to somebody else doing it.

With this basis, something that *would* work under such circumstances is the addition of a cargo feature that uses this build system *instead of* the default one. I would even mark it with `use-alternative-build-system-alpha` to indicate it's early and unstable and may not even be working unless it's kept working by the people who are using it.

That way the maintenance burden is shifted to contributors and there are no guarantees that it keeps working, which allows us to learn what it means to have it and maybe maintain it. Something I can imagine will happen is that as the build system gets patched to support upstream changes, I see and learn how to do that, maybe realising that it is indeed very doable which would ease the transition to being maintained here as well. Something that could also happen is the 'devil in the detail' which shows it's much less feasible, leading to its removal. But that way we could at least find out as this feature is *without any doubt* valuable and should be considered.

Jake-Shadle:
I've now split it so that cmake and cc are supported in parallel, cmake with the regular `zlib-ng` feature and cc with `zlib-ng-no-cmake` that way all of the crates that currently use `zlib-ng` will continue working as before, but crates that don't want to use cmake can expose a feature flag to use the non-cmake feature instead.

The code is much more simple now and no longer does intrinsic/header detection, and rather just turns on all reasonable features. Amusingly, this means that several linux-gnu targets are skipped when building without cmake in CI due to those targets using an ancient (almost 8 years old) version of GCC that doesn't support some AVX-512 flags.

Byron (changes requested):
Thanks a lot for these improvements!

My change request is primarily about assuring this works with `--all-features` as well. Besides that I only skimmed the `cc.rs` addition as a true requires expert knowledge that I clearly don't have. It would certainly be helpful to structure it so that parts that need to change often, based on changes in the `zlib-ng` sources, would be easy to find and to adjust, in case that's already known.

I also very much welcome bikeshedding the name of the feature toggle so that it expresses its volatile nature that 'can break at any time'.

With that said, I even though I think it should be tested in CI, I also wouldn't want to be blocked on it. After all, it should be community supported so (effectively) I don't have to support a whole new build system that I don't know for C sources that I don't know either. Thus far, `zlib-ng` source updates where absolutely painless and just worked which is a feature I really want to retain.

A solution could probably be to have the script emit a special 'warning' on error (instead of failing) which will show up in the CI logs and be visible enough so I can CC @Jake-Shadle when I see it. GHA has such a feature that is based on magic in stdout/stderr.

For alignment, the way I see the progression of this PR, once I am comfortable merging, I'd definitely ask for another reviewer to double-check my sanity 😁, this is big and unconventional, and I do have doubts.

After all, another way of solving this problem is to have a fork of this crate, and projects which need to rid themselves of `cmake` will patch it in, while the maintainer of the project re-applies a patch-queue which could probably be so minimal that it basically never fails (maybe with some of the modifications of this PR to make that easier). And even if this turns out to be the better option which I leave to other maintainers or the second reviewer to decide, I think this work here will help getting there so it's never lost. This disadvantage would be that I'd never learn how hard it really is in practice to maintain the feature, and that would be a loss.

Thanks for bearing with me - but for this crate I have to scrutinise my scrutiny to assure nothing breaks.

zng/cc.rs
```diff
@@ -0,0 +1,441 @@
...
+                if !is_aarch64 {
+                    cfg.mflag("-march=armv8-a+crc", None);
+                    cfg.define("ARM_ASM_HWCAP", None);
+                }
+            }
+
+            // neon
+            cfg.define("ARM_NEON", None);
+
+            // TODO: These intrinsics were only added in gcc 9.4, which is _relatively_
```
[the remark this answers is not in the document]
Jake-Shadle: Possible, sure, it was what it was doing before, but others were concerned about complexity so I removed it. I think it makes sense to have such compile checks in a separate project so that people can add them collaboratively rather than being siloed in a single project, making it easier for other projects to remove cmake as well.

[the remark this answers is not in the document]
Byron: I see - then I think it's fair to update the comment to reflect that this compile-time feature check should come from a separate crate so it's not done here. I also think it makes sense to do that - something I try to avoid is TODO's that probably stay forever - so maybe that should even be a `NOTE` explaining why certain flags aren't set here.

.github/workflows/ci.yml
```diff
@@ -79,6 +79,26 @@ jobs:
...
 
+  cc:
+    runs-on: ubuntu-22.04
+    env:
+      TARGET_TRIPLE: x86_64-unknown-linux-gnu
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          submodules: true
+      - uses: dtolnay/rust-toolchain@stable
```
Byron, line 90: I don't think there is a need for this line. It's should be fine to use the CI provided `rustc`.

.github/workflows/ci.yml
```diff
@@ -79,6 +79,26 @@ jobs:
...
+      # Ensures --all-features builds correctly, the current logic will mean it
+      # uses stock zlib, not cmake nor cc
+      - run: |
+          cargo test --target $TARGET_TRIPLE --all-features
+          cargo run --target $TARGET_TRIPLE --manifest-path systest/Cargo.toml --all-features
+      # Ensures zlib-ng builds and runs, though zlib-ng _could_ change internally
+      # and not use all optimizations available to the CI runner
+      - run: |
```
Byron, line 98: This run block seems to test more zlib-ng-no-cmake features which might fail, even though this job should only fail if `--all-features` isn't working as expected (i.e., use `stock-zlib` then).

Thus I think this block has to be removed.

ci/test.bash
```diff
@@ -31,16 +31,30 @@ fi
...
 $CROSS test --target $TARGET_TRIPLE --no-default-features --features zlib-ng
 $CROSS run --target $TARGET_TRIPLE --manifest-path systest/Cargo.toml --no-default-features --features zlib-ng
-echo === libz-ng-sys build ===
+echo '::endgroup::'
+
+# Note we skip compiling these targets on CI because the gcc version currently used in
+# cross for them is 5.4, ~8 years old at this point, hopefully it will be updated...sometime
+skip_triples=("x86_64-unknown-linux-gnu" "i686-unknown-linux-gnu" "aarch64-unknown-linux-gnu" "arm-unknown-linux-gnueabihf" "s390x-unknown-linux-gnu")
+if [[ -z $CI ]] || ! [[ ${skip_triples[@]} =~ "${TARGET_TRIPLE}" ]]; then
+    echo '::group::=== zlib-ng-no-cmake build ==='
```
Byron, line 44: This test would still halt CI on failure, even though it should produce a warning. If CI halts when this breaks unrelated changes are also held up even though they shouldn't.
The intended workflow is that I CC you when I see such warning.

Byron (changes requested):
Thanks a lot!

One point of the previous review wasn't addressed yet, and when that's fixed, I will ask for another reviewer in Zulip.

Thanks.

.github/workflows/ci.yml
Jake-Shadle: I've made it non-fatal, this way a warning gets printed if a common target would fail that _actually_ uses the intrinsics since the cross versions are too old to support it.

Byron (approved):
Thanks a lot for all the work you put into this!

I asked for [another review](https://rust-lang.zulipchat.com/#narrow/stream/351149-t-libs.2Fcrates/topic/request.20for.20additional.20review.20on.20libz-ng-sys/near/420068455) on Zulip, which if positive would lead to this PR being merged.

zng/cc.rs
```diff
@@ -0,0 +1,470 @@
...
+            write(&[
+                IoSlice::new(begin.as_bytes()),
+                IoSlice::new(end.as_bytes()),
+                IoSlice::new(b"\n"),
+            ]);
+        } else {
+            write(&[IoSlice::new(line.as_bytes()), IoSlice::new(b"\n")]);
+        }
+
+        if get_version {
```
wcampbell0x2a, line 89: To avoid the `contains` call after you find the version.

```suggestion
        if version.is_none() && get_version {
```

ofek:
Just a heads up that this requires a rebase.

ofek:
I just encountered this as well. We have a project with well-documented instructions for building from source and upgrading zip2 (which in turn pulled in `libz-ng-sys`) would break all of our downstream users by now requiring CMake.

Jake-Shadle:
I'm going to close this since it doesn't appear it will be merged, I'll leave the fork up for now in case someone wants to carry if forward, or just use it for reference or something.

ofek:
If you still have time I think it would be good to click open once more because I think there is desire for this to be merged but the rebase happened only a few days ago to resolve the conflicts. I and a few others I know of have a great need for this so it would be nice to give the maintainers a little bit more time to review, please 🙏

ofek:
I understand your perspective, thanks. It's a bummer that so many folks will not use ng/the best implementation because of CMake. I don't have time unfortunately or else I would do this myself.

Jake-Shadle:
I'd be happy to have this merged, was just closing PRs and issues in my backlog that didn't seem to be going anywhere.

Byron:
Great to hear! If a release is desired, I'd need a PR with the version change as I don't think I can merge my own PRs.

ofek:
Is there no way to use this new option with cross? https://github.com/ofek/pyapp/pull/137

I'm guessing the images cross uses have very old compiler versions?

PR #193, opened 2024-05-16
Bump version to 1.1.17

Jake-Shadle:
Normally I would also update the CHANGELOG to add #174, but this project doesn't have one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 name = "libz-sys"
-version = "1.1.16"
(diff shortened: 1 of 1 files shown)


Byron:
I wonder if this should rather be a patch as the feature is experimental and not quite for general consumption yet. Personally, I'd feel better with a patch-bump, but understand why it would be a valid minor bump as well.

Byron:
Thanks a lot! The new release is now available here: https://github.com/rust-lang/libz-sys/releases/tag/1.1.17

PR #196, opened 2024-05-29
cc: Only enable ARM_NEON when in target_feature

wcampbell0x2a:
* Fix armv7-unknown-linux-musleabi and arm-unknown-linux-musleabi by only passing in ARM_NEON if that target is enabled.
* This was only tested without neon, and *not* tested *with* neon support

To enable NEON in the future, I think the following will need to be addressed with adding compiler options:
```
  cargo:warning=                 from src/zlib-ng/arch/arm/slide_hash_neon.c:12:
  cargo:warning=/usr/local/lib/gcc/arm-linux-musleabi/9.2.0/include/arm_neon.h:31:2: error: #error "NEON intrinsics not available with the soft-float ABI.  Please use -mfloat-abi=softfp or -mfloat-abi=hard"
  cargo:warning=   31 | #error "NEON intrinsics not available with the soft-float ABI.  Please use -mfloat-abi=softfp or -mfloat-abi=hard"
```

### DIFF
--- a/zng/cc.rs
+++ b/zng/cc.rs
@@ -196,6 +196,7 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
 
     let target_os = env::var("CARGO_CFG_TARGET_OS").unwrap();
     let arch = env::var("CARGO_CFG_TARGET_ARCH").expect("failed to retrieve target arch");
+    let features = env::var("CARGO_CFG_TARGET_FEATURE").unwrap();
 
     let is_linux_or_android = matches!(target_os.as_str(), "linux" | "android");
     if is_linux_or_android {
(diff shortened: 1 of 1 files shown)


zng/cc.rs
```diff
@@ -311,7 +312,9 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             }
 
             // neon
-            cfg.define("ARM_NEON", None);
+            if features.contains("neon") {
```
Byron, line 315: Even though right now `neon` is unique, I am afraid that in future there will be `neon-ultra` or something like that which might make this a false positive.

Can you try `features.split(",").any(|name| name == "neon")` instead?

Byron (changes requested):
Thank you for this fix - it seems reasonable to conditionally add this feature.

There is just one small robustness change to be done before the merge.
Thanks again.

zng/cc.rs
wcampbell0x2a: Sure, fixed with a62591dc3f3f9cce04f7a02459618c5767243a0e

Byron:
Thanks a lot!

PR #198, opened 2024-07-11
Set DEP_Z_INCLUDE if zlib from vcpkg

sagudev:


Byron (changes requested):
Could you elaborate?
Please also feel free to put the motivation into a comment so future readers find out about it more easily.

sagudev:
> Could you elaborate? Please also feel free to put the motivation into a comment so future readers find out about it more easily.

Well, `DEP_Z_INCLUDE` is already set when zlib is found via pkg-config: https://github.com/rust-lang/libz-sys/blob/50dfd5548d121d4532262827762aebfec7e05cc0/build.rs#L44-L53 so I see no reason why we wouldn't do the same for vcpkg (I somehow expected it already). This is useful for other build script to be able to get include dir of zlib in more cases, instead of doing queries themself.

Byron:
Thanks!

Could you put the motivation/explanation above into the commit message so I can merge?

Thanks again.

Byron (approved):
Thanks a lot for contributing, much appreciated.

PR #199, opened 2024-07-26
Do not check pkg-config when cross-compiling for android

sagudev:
Currently when compiling for android/haiku target we check pkg-config
for zlib then fallback to using shipped zlib,
this can cause problems as pkg-config sometimes returns host's zlib.

This patch makes it to always use shipped zlib instead of pkg-config one (moves `target.contains("android")` check before pkg-config check).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -22,6 +22,14 @@ fn main() {
         return build_zlib_ng(&target, true);
     }
 
+    // All android compilers should come with libz by default, so let's just use
+    // the one already there. Likewise, Haiku always ships with libz, so we can
+    // link to it even when cross-compiling.
+    if target.contains("android") || target.contains("haiku") {
+        println!("cargo:rustc-link-lib=z");
+        return;
+    }
+
     // Don't run pkg-config if we're linking statically (we'll build below) and
     // also don't run pkg-config on FreeBSD/DragonFly. That'll end up printing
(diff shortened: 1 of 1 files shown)


sagudev:
CI has some network issues

Byron (changes requested):
Could you provide a motivation for this change either as commit message, as in-code comment, or a combination of both?

Thank you!

Byron:
Could you spell out everything in the commit message, like I am 3 years old?
The current explanation doesn’t tell me much. Imagine someone who has no background in this.
Thanks again

sagudev:
In case commit message is unclear, you can also edit it yourself (I always `Allow edits by maintainers`).

Byron:
I really meant the 'please explain it like I am 3' because I wouldn't know how to improve it - it's entirely unclear to me, I have no context.

sagudev:
But it is ok now? (There was a force push between https://github.com/rust-lang/libz-sys/pull/199#issuecomment-2253110900 and https://github.com/rust-lang/libz-sys/pull/199#issuecomment-2253289469)

Byron (approved):
Thanks a lot, I understand what's going on now and can confirm that the change makes sense to accomplish this goal (without having tested or otherwise validated it).

PR #206, opened 2024-08-15
Add LIB_SYS_TRY_SHARED ENV var to try to link with shared library

### LINKED ISSUE #201: Link against shared zlib library when cross-compiling
Currently libz-sys is always building a static zlib library when cross-compiling, except on Apple platforms. Can I ask the reason of this? I have a shared zlib library in my rootfs, and pkg-config is working fine. With this config, I have a build failure because of library collision, like this:

`/tmp/rustc0ZqD0F/liblibz_sys-84983a050a121d20.rlib(inflate.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `__stack_chk_guard' which may bind externally can not be used when making a shared object; recompile with -fPIC`

If I remove forcing this static library build, it is building fine:
`
diff --git a/build.rs b/build.rs
index cab160ae6e6f..87e3e8664fd6 100644
--- a/build.rs
+++ b/build.rs
@@ -81,11 +81,11 @@ fn main() {
     //
     // Apple platforms have libz.1.dylib, and it's usually available even when
     // cross compiling (via fat binary or in the target's Xcode SDK)
-    let cross_compiling = target != host;
+    // let cross_compiling = target != host;
     if target.contains("msvc")
         || target.contains("pc-windows-gnu")
         || want_static
-        || (cross_compiling && !target.contains("-apple-"))
+        // || (cross_compiling && !target.contains("-apple-"))
     {
         return build_zlib(&mut cfg, &target);
     }
`

hodrob84:
On cross-compilation libz-sys is always building static zlib library, except for Apple platforms. If target has an installed zlib shared library, this is problem, because:

  1. the binary is increased, because we are including a static zlib as well
  2. it may cause build issues if the pkg-config sets the linker arguments, like this:

     /tmp/rustc0ZqD0F/liblibz_sys-84983a050a121d20.rlib(inflate.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol '__stack_chk_guard' which may bind externally can not be used when making a shared object; recompile with -fPIC

To workaround the issue, add a LIB_SYS_TRY_SHARED=1 ENV var (similarly to LIBZ_SYS_STATIC=1) to force the use of the shared library, recognized by the pkg-config.

fixes #201

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -82,10 +82,12 @@ fn main() {
     // Apple platforms have libz.1.dylib, and it's usually available even when
     // cross compiling (via fat binary or in the target's Xcode SDK)
     let cross_compiling = target != host;
+    let try_link_with_shared =
+        env::var("LIB_SYS_TRY_SHARED").unwrap_or(String::new()) == "1";
     if target.contains("msvc")
         || target.contains("pc-windows-gnu")
         || want_static
-        || (cross_compiling && !target.contains("-apple-"))
+        || !try_link_with_shared && (cross_compiling && !target.contains("-apple-"))
     {
         return build_zlib(&mut cfg, &target);
     }


Byron:
Thanks a lot for tackling this.

@jongiddy I feel that `libz-sys` gets more complex by adding patch upon patch, and hoped you could involve yourself here more as well. Maybe you could help with making a case for using environment variables to alter build results, or for using Cargo Features. [This is my related comment](https://github.com/rust-lang/libz-sys/issues/201#issuecomment-2292915190), to allow for backward compatible fixes for those who know about it.

Thanks everyone for your help.

jongiddy:
Neither option is ideal.

Using an environment variable means there is a hidden dependency that affects the build. When anyone needs support, it is an extra relevant piece of information they need to supply. And it is easy to not be aware whether it is set or not.

Using a feature is painful when a transitive dependency is deeply nested, and `libz-sys` is typically deeply nested. You either need to pass the feature setting down through the intermediate dependencies or include the dependency in the top-level crate, even though it is not used directly.

Pragmatically, since there is already an environment variable for controlling static vs dynamic, I'd accept using an environment variable here. However, I suggest that the existing variable is reused, so that setting `LIBZ_SYS_STATIC=0` always forces dynamic linking, even for the supposedly non-supported targets (e.g. `target.contains("msvc")`).

wezm:
FWIW there is precedent in other `-sys` crates for disabling building of vendored C code/forcing use of system library through an env var:

- libgit2-sys: "LIBGIT2_NO_VENDOR": "1",
- libssh2-sys: "LIBSSH2_SYS_USE_PKG_CONFIG": "1",
- openssl-sys: "OPENSSL_NO_VENDOR": "1",
- pcre2-sys: "PCRE2_SYS_STATIC": "0",
- rustonig-sys: "RUSTONIG_SYSTEM_LIBONIG": "1",
- zstd-sys: "ZSTD_SYS_USE_PKG_CONFIG": "1",
- libsqlite3-sys: "LIBSQLITE3_SYS_USE_PKG_CONFIG": "1",

Byron:
Thanks a lot for the summary, it's good to know the standards.

This PR is at a stage where the question remains if `LIBZ_SYS_STATIC=0` can be used instead of introducing a new environment variable. On top of that, I think there should be a section in the crate-level docs to document this behaviour.

jongiddy:
> * pcre2-sys: "PCRE2_SYS_STATIC": "0",

This one is interesting. Matches the [proposed one variable model](https://crates.io/crates/pcre2-sys/0.2.9) with 0 meaning force dynamic and 1 meaning force static.

tsjk:
I'm hitting this when trying to cross-compile a project.
I have zlib1g-dev:arm64 installed in the host, but compilation fails because the compiler is called in a bad way - probably because it tries to statically link against a foreign architecture library (I see `-lz`).
The zlib-dependency is deep, so it´s not in the project's `Cargo.toml`.
Is there some way to work this around?

Byron:
I think this patch is stuck on [this question](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-2325610571) along with the implementation lacking discoverability as there are no docs at all.
Maybe the best discoverability could be achieved by putting typical compiler errors into the library docs, along with how to use certain environment variables to fix it.

tsjk:
> I think this patch is stuck on [this question](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-2325610571) along with the implementation lacking discoverability as there are no docs at all. Maybe the best discoverability could be achieved by putting typical compiler errors into the library docs, along with how to use certain environment variables to fix it.

Is there something I can do in the meanwhile to be able to cross-compile the project I'm looking at?

Byron:
One could tell Cargo to use the patched version of the dependency (with `[patch."crates-io"]`) to make the environment variable override available.
Alternatively, one could pick up this PR and push it to completion.

tsjk:
> One could tell Cargo to use the patched version of the dependency (with `[patch."crates-io"]`) to make the environment variable override available. Alternatively, one could pick up this PR and push it to completion.

In my case this does not seem to help.
I did
```
ENV target_host=aarch64-linux-gnu \
    target_host_rust=aarch64-unknown-linux-gnu \
    AR=${target_host}-ar \
    AS=${target_host}-as \
    CC=${target_host}-gcc \
    CXX=${target_host}-g++ \
    LD=${target_host}-ld \
    RANLIB="${target_host}-ranlib" \
    STRIP=${target_host}-strip \
    PKG_CONFIG_PATH="/usr/${target_host}/lib/pkgconfig" \
    PKG_CONFIG_SYSROOT_DIR="/usr/${target_host}" \
    LIB_SYS_TRY_SHARED=1
mkdir -p .cargo && \
                echo "[target.${target_host_rust}]" > .cargo/config.toml && \
                echo "linker = \"${target_host}-gcc\"" >> .cargo/config.toml && \
                echo "ar = \"${target_host}-ar\"" >> .cargo/config.toml && \
                echo "" >> .cargo/config.toml && \
                echo "[patch.crates-io]" >> .cargo/config.toml && \
                echo "libz-sys = { path = '/tmp/libz-sys' }" >> .cargo/config.toml && \
                cargo build --target="${target_host_rust}" --release
```

with zlib manually installed, just in case - and this in a Debian image:
```
__ZLIB_VERSION='1.2.13' && \
      ( cd /tmp && \
          wget -O /tmp/zlib-${__ZLIB_VERSION}.tar.gz https://zlib.net/fossils/zlib-${__ZLIB_VERSION}.tar.gz && \
          tar xf zlib-${__ZLIB_VERSION}.tar.gz && \
          rm zlib-${__ZLIB_VERSION}.tar.gz )
```

but it seems to not understand, and the aarch64 compiler is called with invalid flags
```
The following warnings were emitted during compilation:                                                                                                                                                                                                                                                                                                                                                 
warning: libz-sys@1.1.20: aarch64-linux-gnu-gcc: error: unrecognized command-line option ‘-m64’
warning: libz-sys@1.1.20: ToolExecError: Command "aarch64-linux-gnu-gcc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "src/zlib" "-fvisibility=hidden" "-DSTDC" "-D_LARGEFILE64
_SOURCE" "-D_POSIX_SOURCE" "-o" "/boltz-backend/boltzr/target/release/build/libz-sys-2e1975913bc64b22/out/lib/0dc752f03a07a721-adler32.o" "-c" "src/zlib/adler32.c" with args aarch64-linux-gnu-gcc did not execute successfully (status code exit status: 1).cargo:warning=aarch64-linux-gnu-gcc: error: unrecognized command-line option ‘-m64’

**Comment by Byron** [comment]
Maybe a different issue? The local override can also point to a local checkout of the crate, which might help to make adjustments to `build.rs` for a fix. Maybe something that can eventually be fully understood and contributed.

**Comment by tsjk** [comment]
Yes, this was related to something else. I managed to cross-compile by overriding the dependency. Thanks. :)

**Comment by polarathene** [comment]
@tsjk you can more easily cross-compile via Zig (_with [`cargo-zigbuild`](https://github.com/rust-cross/cargo-zigbuild) for Rust support_).

Here's a full example within a Docker container (Fedora 42) as an example, but you should be able to adapt to your equivalent environment rather easily:

```bash
# Fedora container for a reproducible build environment, running on an amd64 / x86_64 host:
$ docker run --rm -it --workdir /example fedora:42

# Install base deps:
# Boltz specific deps needed to compile from host arch: npm + protobuf-compiler
dnf install -yq gcc git-core npm protobuf-compiler rustup zig

# Setup Rust with Zig:
rustup-init -y --profile minimal --default-toolchain stable
export PATH="/root/.cargo/bin:${PATH}"
cargo install cargo-zigbuild

# Add cross-compilation support (libs needed for linking + arm64 target):
# NOTE: `openssl-sys` additionally needs to parse headers and since the host vs target arch differs,
# you will need to use `OPENSSL_DIR` or `PKG_CONFIG_SYSROOT_DIR` ENV.
export CROSS_ROOT_ARM64=/opt/sysroot/aarch64-gnu
dnf --installroot ${CROSS_ROOT_ARM64} --use-host-config --setopt=install_weak_deps=0 --forcearch=aarch64 \
  install -yq libpq-devel openssl-devel
rustup target add aarch64-unknown-linux-gnu

# Grab the project source:
git clone --recurse-submodules https://github.com/BoltzExchange/boltz-backend .

# Rust build relies upon files generated from `node_modules`:
# NOTE: Using lock file fails due to outdated `node-gyp@8.4.1` requirement, where Python 3.13 no longer providing a distutils module
npm install --no-package-lock

# Now you can build the rust project for the foreign arch:
# - `PKG_CONFIG_SYSROOT_DIR` is only needed for cross-compilation for `openssl`,
#   alternatively skip the need for `pkg-config` via `OPENSSL_DIR="${CROSS_ROOT_ARM64}/usr"`
# - `RUSTFLAGS` must be set to the aarch64 sysroot for correctly linking to the arm64 `.so` files
cd boltzr
PKG_CONFIG_SYSROOT_DIR="${CROSS_ROOT_ARM64}" RUSTFLAGS="-L ${CROSS_ROOT_ARM64}/usr/lib64" cargo zigbuild --release --target aarch64-unknown-linux-gnu
```

I think that is simpler than using Cargo to cross-compile without Zig 😅

The PR was not used in my example for a successful build via Zig, so it would have built from source successfully? (_**EDIT:** Seems that it was a build-dep, whereas your linked libz for OpenSSL is runtime, thus no conflict_)

---

**Verification:**

`libz` is linked:

```console
# Required as not running via `chroot` or via an aarch64 container/host:
# - `ld-linux-aarch64.so.1 --list` is the equivalent to `ldd` (which is often a shell script wrapper).
# - `LD_LIBRARY_PATH` to add libs from the aarch64 sysroot to the search path for resolving libs.
$ LD_LIBRARY_PATH=/opt/sysroot/aarch64-gnu/usr/lib64 /opt/sysroot/aarch64-gnu/lib/ld-linux-aarch64.so.1 --list target/aarch64*/release/boltzr

        linux-vdso.so.1 (0x0000400000810000)
        libpq.so.5 => /opt/sysroot/aarch64-gnu/usr/lib64/libpq.so.5 (0x0000400003d30000)
        libssl.so.3 => /opt/sysroot/aarch64-gnu/usr/lib64/libssl.so.3 (0x0000400003da0000)
        libcrypto.so.3 => /opt/sysroot/aarch64-gnu/usr/lib64/libcrypto.so.3 (0x0000400003e90000)
        libm.so.6 => /opt/sysroot/aarch64-gnu/usr/lib64/libm.so.6 (0x0000400004290000)
        libpthread.so.0 => /opt/sysroot/aarch64-gnu/usr/lib64/libpthread.so.0 (0x0000400004360000)
        libc.so.6 => /opt/sysroot/aarch64-gnu/usr/lib64/libc.so.6 (0x0000400004390000)
        libdl.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/libdl.so.2 (0x0000400004560000)
        /lib/ld-linux-aarch64.so.1 => /opt/sysroot/aarch64-gnu/lib/ld-linux-aarch64.so.1 (0x00007cf124eb0000)
        libgssapi_krb5.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/libgssapi_krb5.so.2 (0x0000400004590000)
        libldap.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/libldap.so.2 (0x0000400004600000)
        libz.so.1 => /opt/sysroot/aarch64-gnu/usr/lib64/libz.so.1 (0x0000400004690000)
        libkrb5.so.3 => /opt/sysroot/aarch64-gnu/usr/lib64/libkrb5.so.3 (0x00004000046d0000)
        libk5crypto.so.3 => /opt/sysroot/aarch64-gnu/usr/lib64/libk5crypto.so.3 (0x00004000047b0000)
        libcom_err.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/libcom_err.so.2 (0x00004000047f0000)
        libkrb5support.so.0 => /opt/sysroot/aarch64-gnu/usr/lib64/libkrb5support.so.0 (0x0000400004820000)
        libkeyutils.so.1 => /opt/sysroot/aarch64-gnu/usr/lib64/libkeyutils.so.1 (0x0000400004850000)
        libresolv.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/libresolv.so.2 (0x0000400004880000)
        liblber.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/liblber.so.2 (0x00004000048b0000)
        libevent-2.1.so.7 => /opt/sysroot/aarch64-gnu/usr/lib64/libevent-2.1.so.7 (0x00004000048f0000)
        libsasl2.so.3 => /opt/sysroot/aarch64-gnu/usr/lib64/libsasl2.so.3 (0x0000400004970000)
        libselinux.so.1 => /opt/sysroot/aarch64-gnu/usr/lib64/libselinux.so.1 (0x00004000049b0000)
        libcrypt.so.2 => /opt/sysroot/aarch64-gnu/usr/lib64/libcrypt.so.2 (0x0000400004a00000)
        libpcre2-8.so.0 => /opt/sysroot/aarch64-gnu/usr/lib64/libpcre2-8.so.0 (0x0000400004a50000)
```

But as `patchelf` shows, not directly, it is via the linked OpenSSL dep:

```console
# Alternatively... chroot in:
$ cp target/aarch64*/release/boltzr /opt/sysroot/aarch64-gnu/tmp/boltzr
$ dnf --installroot /opt/sysroot/aarch64-gnu --use-host-config --forcearch=aarch64 install patchelf
$ chroot /opt/sysroot/aarch64-gnu

# Direct deps linked (zlib from ldd was implicit from openssl link):
$ patchelf --print-needed /tmp/boltzr
libpq.so.5
libssl.so.3
libcrypto.so.3
libm.so.6
libpthread.so.0
libc.so.6
libdl.so.2
ld-linux-aarch64.so.1

$ patchelf --print-needed /usr/lib64/libcrypto.so.3
libz.so.1
libc.so.6
ld-linux-aarch64.so.1

# `libcrypto` is the only dep requiring `libz`:
$ exit
$ cargo install lddtree
$ lddtree target/aarch64*/release/boltzr /opt/sysroot/aarch64-gnu | grep -B1 libz
# Output omitted

# Here we can see that boltzr project uses `libz-sys` crate as a build-dep:
$ cargo tree -i libz-sys
libz-sys v1.1.20
└── libgit2-sys v0.18.0+1.9.0
    └── git2 v0.20.0
        └── built v0.8.0
            [build-dependencies]
            └── boltzr v3.11.0 (/example/boltzr)
```

So in your case `libz-sys` is specifically built as a build-dep.

build.rs
```diff
@@ -82,10 +82,12 @@ fn main() {
     // Apple platforms have libz.1.dylib, and it's usually available even when
     // cross compiling (via fat binary or in the target's Xcode SDK)
     let cross_compiling = target != host;
+    let try_link_with_shared =
+        env::var("LIB_SYS_TRY_SHARED").unwrap_or(String::new()) == "1";
     if target.contains("msvc")
         || target.contains("pc-windows-gnu")
         || want_static
-        || (cross_compiling && !target.contains("-apple-"))
+        || !try_link_with_shared && (cross_compiling && !target.contains("-apple-"))
```
polarathene: This line is no longer relevant, and should be left alone as it is now currently:

https://github.com/rust-lang/libz-sys/blob/80c597a07f5e5f0dff0e98c03d48a77845de9f6e/build.rs#L81-L86

polarathene (changes requested):
There you go @Byron , I haven't got time atm to give it too much thought but the method I've suggested in the review comment should be able to define `want_static` variable correctly?

Presently there is no way to opt-out of the static feature, so the equivalent ENV is passed to the three possible states, where if it's not `0` or `1` the `static` feature is used as a fallback.

So basically `false` is the expected default, the `static` feature or ENV can flip that to `true`, but if an upstream crate has set the `static` feature the ENV now also supports reverting that decision 👍

Byron:
@polarathene Thanks so much for chiming in!

What do you think about leaving this mega-thread behind, and starting a new PR?
There I'd love to *start* with documenting the current behaviour of `build.rs` in `lib.rs`, along with a careful refactor for clarity where needed, akin to the code shown in one of your comments.

This we merge as clear win of the (mostly?) undocumented status quo. And from there we iterate and make adjustments, both to the newly added documentation and the code, for mostly self-descriptive changes that solve specific problems.
It's probably not too hard, it's just that I think we are knee-deep in tech-debt that makes working on this unnecessarily hard, with high stakes in terms of blast radius for newly introduced bugs.

So, probably needs some dedication, but I'd be with you with timely reviews and early feedback if desired.
Thanks for your consideration.

polarathene:
> What do you think about leaving this mega-thread behind, and starting a new PR?

Yeah that'd be better, the new PR can reference this one for context.

> There I'd love to _start_ with documenting the current behaviour of `build.rs` in `lib.rs`, along with a careful refactor for clarity where needed, akin to the code shown in one of your comments.

While I can open a new PR with the suggested approach, I don't presently have time to context switch to address the other wants. Possibly in December.

> It's probably not too hard, it's just that I think we are knee-deep in tech-debt that makes working on this unnecessarily hard, with high stakes in terms of blast radius for newly introduced bugs.

The logic involved here doesn't appear to be too risky, we are only interested in adjusting how `want_static` is enabled.
- Presently the `static` feature opt-in supersedes the equivalent ENV support `LIBZ_SYS_STATIC=1`, which is a non-concern as both only support toggling `true`.
- Desired change swaps the order of feature vs ENV, with the added functionality to explicitly opt-out of static via `LIBZ_SYS_STATIC=0`, where we use an `Option` to wrap either bool as `Some(_)` otherwise `None` will fallback to checking the `static` feature opt-in, which if not used results in `false` as was previously the case.

So basically only difference is `LIBZ_SYS_STATIC=0` is supported when it wasn't before and it has priority for setting `want_static` over the crate `static` feature. I don't see any breaking change or bug concerns from such affecting any existing users.

> So, probably needs some dedication, but I'd be with you with timely reviews and early feedback if desired.
> Thanks for your consideration.

I appreciate that, I would like to land the improvement so I'll raise a PR. I just won't be able to dedicate much time this month if the PR requires more thorough assessment.

Byron:
Thanks so much @polarathene!

Thanks also for spelling out the logic for me, I agree that this seems like the way to go without affecting any of the existing users.

I will take a look at the PR soon and possibly do changes myself if I feel fit enough to do so.

Byron:
I am trying to get the ball rolling again by also presenting an alternative https://github.com/rust-lang/libz-sys/pull/275 , without stating a preference. Maybe it's possible to decide on steps forward - please feel free to chime in.

PR #207, opened 2024-08-23
Add minimal-versions CI

NobodyXu:
Help discover bugs in version requ
irements

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,3 +144,27 @@ jobs:
           - s390x-unknown-linux-gnu
           - x86_64-unknown-linux-gnu
           - x86_64-unknown-linux-musl
+
+  minimal-versions:
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: true
+
+      - run: |
+          rustup toolchain install nightly --profile minimal --no-self-update
+          cargo +nightly update -Zminimal-versions
+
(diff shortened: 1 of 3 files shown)


NobodyXu:
cc @Byron seems like libz-sys is ok?

I'm not sure if I'm running with the right feature set

Byron:
Even though it says it uses minimal versions, I am missing `cargo update` with minimal versions set. Right now it would use Cargo.lock or the latest available, *I think*.

NobodyXu:
Thanks, I was braindead when writing this 😂

.github/workflows/ci.yml
```diff
@@ -144,3 +144,22 @@ jobs:
...
           - x86_64-unknown-linux-musl
+
+  minimal-versions:
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: true
+
+      - run: cargo build
```
NobodyXu, line 155:
```suggestion
      - run: |
          rustup toolchain install nightly --profile minimal --no-self-update
          cargo +nightly update -Zminimal-versions

      - run: cargo build
```

Byron:
I tried to get the actual `Cargo-zng.toml` file tested as well, but on Windows it's failing as it appears to be unable to find visual studio related environment variables, and I couldn't figure out why.
So I just left it untested, but applied the expected fix anyway.

Once CI is green I can create a new release and then fix `flate2`, which ultimately fixes `gitoxide`.

Byron (approved):

PR #208, opened 2024-08-23
Fix unknown-cfg warning in libz-ng

NobodyXu:


Byron (approved):
Thanks a lot! I noticed these too, but ignored them till now.

I wonder if they could be fixed by removing them, instead of suppressing them. Probably that's more work though, so I am happy to go with this.

CC @jongiddy - I think merging this is an improvement, but I wonder if you have opinions on fixing the underlying issue. Is it fixable even?

PR #209, opened 2024-08-28
Fix cargo:warning syntax in build.rs

kornelski:
The warning wasn't actually printed by Cargo


Byron (approved):
That's a great catch, thank you!

PR #211, opened 2024-09-03
Update zlib-ng library to 2.2.1

JakubOnderka:


zng/cc.rs
```diff
@@ -137,13 +133,25 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
...
             "insert_string_roll",
-            "slide_hash",
             "trees",
             "uncompr",
             "zutil",
         ],
     );
 
+    cfg.append(
+        Some("arch/generic"),
```
Byron, line 143: Can you fill me in a little why this was necessary? Did these files move?

zng/cc.rs
```diff
@@ -238,7 +246,7 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
 
             // SSE4.2
             cfg.define("X86_SSE42", None);
-            cfg.append(Some("arch/x86"), &["adler32_sse42", "insert_string_sse42"]);
```
Byron, line 241: What's the background to removing this?

Byron (commented):
Thanks a lot for the upgrade! I added a few general questions, and also cordially invite @Jake-Shadle to take a look as the original creator.

zng/cc.rs
JakubOnderka: Yes, these files moves from `/` to `/arch/generic`, see this change https://github.com/zlib-ng/zlib-ng/compare/2.1.x...2.2.x#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

zng/cc.rs
JakubOnderka: `insert_string_sse42` doesn't exists anymore.

Byron (approved):
Thanks a lot!

PR #217, opened 2024-09-07
Also `grep -v` out `.cargo_vcs_info.json` in `cargo-zng`

### LINKED ISSUE #214: In `cargo-zng` on a recent toolchain, `tar` fails on `.cargo_vcs_info.json`
When I running `./cargo-zng` on `cargo` 1.81.0, I get:

```text
tar: .cargo_vcs_info.json: Cannot stat: No such file or directory
tar: Error exit delayed from previous errors.
```

I first observed this on a macOS 14.6 system, which if I recall correctly I had run `rustup update` on shortly before. I had thought the cause had something to do with macOS before, but now I believe it is triggered by some change in 1.81.0, because I have since observed it on two Ubuntu system and Windows, all of which I had run `rustup update` on between not seeing it and seeing it.

This happens in the `tar` pipeline, so it occurs whether or not there is a subcommand. I've verified that it happens both when the full command is `./cargo-zng` and when it is `./cargo-zng build`. The latter seems still to build without errors on systems that did not have errors before, including macOS.

I found this when testing #213 but it occurs, at least on the test machines I am using, with or without those changes.

Running `bash -x cargo-zng` on macOS produces this output, followed by the output of `cargo` itself which I have omitted here:

```text
+ set -eu
+ trap 'rm -rf "$tempdir"' 0 INT
++ mktemp -d
+ tempdir=/var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51
+ cargo package -l --allow-dirty
+ grep -v '^Cargo\.toml\.orig$'
+ tar --files-from=- -cf -
+ tar -C /var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51 -xf -
tar: .cargo_vcs_info.json: Cannot stat: No such file or directory
tar: Error exit delayed from previous errors.
+ cp Cargo-zng.toml /var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51/Cargo.toml
+ cp -a systest /var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51/systest
+ mv /var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51/systest/Cargo-zng.toml /var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51/systest/Cargo.toml
+ cd /var/folders/8q/fntngxw15tz1vjj_93ms_jf00000gq/T/tmp.iMaAxAzP51
+ cargo
```

EliahKagan:
Fixes #214

In the `cargo-zng` script, this has the `grep -v` command that filters out `Cargo.toml.orig` also filiter out `.cargo_vcs_info.json`.

It also refactors the command to use fixed string (`-F`) and match whole lines (`-x`), to eliminate the `^` and `$` anchors and `\` escapes from the pattern(s).

This is working for `./cargo-zng` and `./cargo-zng build` on Ubuntu 24.04 LTS, macOS 14.6, and Windows 10 in Git Bash.

### DIFF
--- a/cargo-zng
+++ b/cargo-zng
@@ -4,7 +4,7 @@ tempdir="$(mktemp -d)"
 trap 'rm -rf "$tempdir"' 0 INT
 cargo package -l --allow-dirty |
     tr '\\' '/' |
-    grep -v '^Cargo\.toml\.orig$' |
+    grep -vxF -e Cargo.toml.orig -e .cargo_vcs_info.json |
     tar --files-from=- -cf - |
     tar -C "$tempdir" -xf -
 cp Cargo-zng.toml "$tempdir/Cargo.toml"


Byron (approved):
Thanks a lot, much appreciated!

I particularly like how the `grep` line has been improved, I had no idea it could do that!.

PR #218, opened 2024-09-08
Recognize `RISCV_WITH_RVV` env var on RISC-V to set `WITH_RVV` cmake var

### LINKED ISSUE #200: zlib-ng: Illegal instruction on riscv64gc-unknown-linux-gnu
This strongly resembles #148 (https://github.com/Byron/gitoxide/issues/955) but affects riscv64 rather than x86_64, seems to happen every time rather than sporadically, and may be a case of a known zlib-ng bug, https://github.com/zlib-ng/zlib-ng/issues/1705.

#### What happens

Both `dev` and `release` builds are affected. On a 64-bit RISC-V machine running Ubuntu 24.04 LTS, running `gix clone` begins the download but is always terminated with `SIGILL` (as indicated by the message text and as can be inferred from the exit status):

```text
ubuntu@riscv:~/src$ gix clone git@github.com:Byron/gitoxide.git
Illegal instruction (core dumped)              1 steps [===   ===   ===   ===   ===   ===   ===   ===   ===   ===   ===]
ubuntu@riscv:~/src[132]$ ects 39.1K/48.3K objects [81%] [=================================================>------------]
ubuntu@riscv:~/src[132]$                    0B  |0B/s| [===   ===   ===   ===   ===   ===   ===   ===   ===   ===   ===]
```

Passing `--trace` does not help, since it doesn't get far enough to print the traced output, and since the program is immediately being terminated by the system, rather than hitting a panic and being able to unwind. The prompt printed afterwards is interleaved with the displayed progress due to the sudden way the program was terminated. (The displayed exit status, here 132, is [part of the prompt](https://github.com/EliahKagan/pipestatus).)

#### Getting a backtrace using `gdb`

Running it in a debugger and printing a backtrace after the crash provides far more information:

```text
ubuntu@riscv:~/repos/gitoxide (main =)[1]$ file target/debug/gix
target/debug/gix: ELF 64-bit LSB pie executable, UCB RISC-V, RVC, double-float ABI, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-riscv64-lp64d.so.1, BuildID[sha1]=d4b06d7d749d472b1a023c6526c5102b1c885ce8, for GNU/Linux 4.15.0, with debug_info, not stripped
ubuntu@riscv:~/repos/gitoxide (main =)$ gdb target/debug/gix
GNU gdb (Ubuntu 15.0.50.20240403-0ubuntu1) 15.0.50.20240403-git
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "riscv64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from target/debug/gix...
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/ubuntu/repos/gitoxide/target/debug/gix.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) run --trace clone git@github.com:EliahKagan/gitoxide.git
Starting program: /home/ubuntu/repos/gitoxide/target/debug/gix --trace clone git@github.com:EliahKagan/gitoxide.git
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/riscv64-linux-gnu/libthread_db.so.1".
[New Thread 0x3ff78a9f20 (LWP 32276)]
[Detaching after vfork from child process 32278]
[New Thread 0x3ff76a5f20 (LWP 32277)]
[New Thread 0x3ff749bf20 (LWP 32279)]
[Detaching after vfork from child process 32280]
[New Thread 0x3ff7294f20 (LWP 32281)]
[Thread 0x3ff76a5f20 (LWP 32277) exited]
 receiving pack                               1 steps [ ===   ===   ===   ===   ===   ===   ===   ===   ===   ===   ===]
Thread 1 "gix" received signal SIGILL, Illegal instruction.=======================>------------------------------------]
0x0000002aab8abd18 in adler32_rvv_impl (adler=42, /s| [ ===   ===   ===   ===   ===   ===   ===   ===   ===   ===   ===]
    dst=0x500000008 <error: Cannot access memory at address 0x500000008>,
    src=0x2aab8ade4e <load_64_bits+54> "\2037\004\376\003'D\375\263\227", <incomplete sequence \347>,
    len=274877754864, COPY=-1405147728)
    at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libz-ng-sys-1.1.15/src/zlib-ng/arch/riscv/adler32_rvv.c:15
15      static inline uint32_t adler32_rvv_impl(uint32_t adler, uint8_t* restrict dst, const uint8_t *src, size_t len, int COPY) {
(gdb)
(gdb)
(gdb) bt
#0  0x0000002aab8abd18 in adler32_rvv_impl (adler=42,
    dst=0x500000008 <error: Cannot access memory at address 0x500000008>,
    src=0x2aab8ade4e <load_64_bits+54> "\2037\004\376\003'D\375\263\227", <incomplete sequence \347>,
    len=274877754864, COPY=-1405147728)
    at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libz-ng-sys-1.1.15/src/zlib-ng/arch/riscv/adler32_rvv.c:15
#1  0x0000002aab8ad038 in adler32_rvv (adler=1, buf=0x3ffffdb4e7 "\n", len=1)
    at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libz-ng-sys-1.1.15/src/zlib-ng/arch/riscv/adler32_rvv.c:129
#2  0x0000002aab8a36a8 in inf_chksum (strm=0x2aac41b000, src=0x3ffffdb4e7 "\n", len=1)
    at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libz-ng-sys-1.1.15/src/zlib-ng/inflate.c:47
#3  0x0000002aab8a65fa in zng_inflate (strm=0x2aac41b000, flush=0)
    at /home/ubuntu/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libz-ng-sys-1.1.15/src/zlib-ng/inflate.c:1055
#4  0x0000002aab899052 in flate2::ffi::c::{impl#10}::decompress (self=0x2aac3f4b18, input=..., output=...,
    flush=flate2::mem::FlushDecompress::None) at src/ffi/c.rs:252
#5  0x0000002aab89b124 in flate2::mem::Decompress::decompress (self=0x2aac3f4b18, input=..., output=...,
    flush=flate2::mem::FlushDecompress::None) at src/mem.rs:452
#6  0x0000002aab1f7120 in gix_features::zlib::stream::inflate::read<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>> (rd=0x3ffffdd9b0, state=0x2aac3f4b18, dst=...) at gix-features/src/zlib/stream/inflate.rs:20
#7  0x0000002aab2ca7e0 in gix_pack::data::input::bytes_to_entries::{impl#7}::read<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>> (self=0x3ffffdd9b0, into=...) at gix-pack/src/data/input/bytes_to_entries.rs:299
#8  0x0000002aab216c2a in std::io::Read::read_buf::{closure#0}<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>> (b=...)
--Type <RET> for more, q to quit, c to continue without paging--c
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/mod.rs:973
#9  0x0000002aab210806 in std::io::default_read_buf<std::io::Read::read_buf::{closure_env#0}<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>>> (read=...,
    cursor=...) at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/mod.rs:574
#10 0x0000002aab2c833e in std::io::Read::read_buf<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>> (self=0x3ffffdd9b0, buf=...)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/mod.rs:973
#11 0x0000002aab2b61b6 in std::io::copy::stack_buffer_copy<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>, std::io::util::Sink> (reader=0x3ffffdd9b0,
    writer=0x3ffffdda27) at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/copy.rs:278
#12 0x0000002aab3ee0b4 in std::io::copy::{impl#4}::copy_from<std::io::util::Sink, gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>> (self=0x3ffffdda27,
    reader=0x3ffffdd9b0) at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/copy.rs:202
#13 0x0000002aab2b51d2 in std::io::copy::generic_copy<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>, std::io::util::Sink> (reader=0x3ffffdd9b0,
    writer=0x3ffffdda27) at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/copy.rs:89
#14 0x0000002aab359dee in std::sys::pal::unix::kernel_copy::{impl#1}::copy<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>, std::io::util::Sink> (self=...)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/sys/pal/unix/kernel_copy.rs:182
#15 0x0000002aab2b6418 in std::sys::pal::unix::kernel_copy::copy_spec<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>, std::io::util::Sink> (
    read=<optimized out>, write=<optimized out>)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/sys/pal/unix/kernel_copy.rs:76
#16 std::io::copy::copy<gix_pack::data::input::bytes_to_entries::DecompressRead<gix_pack::data::input::bytes_to_entries::PassThrough<&mut std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>, alloc::vec::Vec<u8, alloc::alloc::Global>>>, std::io::util::Sink> (reader=0x3ffffdd9b0, writer=0x3ffffdda27)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/copy.rs:68
#17 0x0000002aab2c9258 in gix_pack::data::input::bytes_to_entries::BytesToEntriesIter<std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>>::next_inner<std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>> (self=0x2aac3f4a30) at gix-pack/src/data/input/bytes_to_entries.rs:117
#18 0x0000002aab2c7f6c in gix_pack::data::input::bytes_to_entries::{impl#2}::next<std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>> (self=0x2aac3f4a30)
    at gix-pack/src/data/input/bytes_to_entries.rs:215
#19 0x0000002aab3ba938 in gix_pack::data::input::lookup_ref_delta_objects::{impl#1}::next<gix_pack::data::input::bytes_to_entries::BytesToEntriesIter<std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>>, alloc::boxed::Box<gix_odb::Cache<gix_odb::store_impls::dynamic::Handle<alloc::sync::Arc<gix_odb::Store, alloc::alloc::Global>>>, alloc::alloc::Global>> (self=0x2aac3f4a30) at gix-pack/src/data/input/lookup_ref_delta_objects.rs:90
#20 0x0000002aab3af36e in core::iter::adapters::peekable::{impl#1}::next<gix_pack::data::input::lookup_ref_delta_objects::LookupRefDeltaObjectsIter<gix_pack::data::input::bytes_to_entries::BytesToEntriesIter<std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>>, alloc::boxed::Box<gix_odb::Cache<gix_odb::store_impls::dynamic::Handle<alloc::sync::Arc<gix_odb::Store, alloc::alloc::Global>>>, alloc::alloc::Global>>> (self=0x2aac3f4a30)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/iter/adapters/peekable.rs:40
#21 0x0000002aab3edca2 in gix_pack::data::input::entries_to_bytes::{impl#1}::next<gix_pack::data::input::lookup_ref_delta_objects::LookupRefDeltaObjectsIter<gix_pack::data::input::bytes_to_entries::BytesToEntriesIter<std::io::buffered::bufreader::BufReader<gix_features::interrupt::Read<gix_features::progress::Read<&mut dyn std::io::BufRead, prodash::progress::utils::ThroughputOnDrop<prodash::traits::BoxedDynNestedProgress>>>>>, alloc::boxed::Box<gix_odb::Cache<gix_odb::store_impls::dynamic::Handle<alloc::sync::Arc<gix_odb::Store, alloc::alloc::Global>>>, alloc::alloc::Global>>, gix_pack::bundle::write::types::LockWriter> (self=0x2aac3f4a30) at gix-pack/src/data/input/entries_to_bytes.rs:130
#22 0x0000002aab23ea00 in alloc::boxed::{impl#42}::next<dyn core::iter::traits::iterator::Iterator<Item=core::result::Result<gix_pack::data::input::Entry, gix_pack::data::input::types::Error>>, alloc::alloc::Global> (self=0x3ffffe0118)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/alloc/src/boxed.rs:1997
#23 0x0000002aab20d706 in core::iter::traits::iterator::{impl#0}::next<dyn core::iter::traits::iterator::Iterator<Item=core::result::Result<gix_pack::data::input::Entry, gix_pack::data::input::types::Error>>> (self=0x3ffffdf498)
    at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/iter/traits/iterator.rs:4109
#24 0x0000002aab3cb3f8 in gix_pack::index::File::write_data_iter_to_stream<gix_pack::bundle::write::{impl#1}::inner_write::{closure_env#0}<&std::path::Path>, fn(core::ops::range::Range<u64>, &memmap2::Mmap) -> core::option::Option<&[u8]>, memmap2::Mmap> (version=gix_pack::index::Version::V2, make_resolver=..., entries=..., thread_limit=...,
    root_progress=..., out=..., should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>,
    pack_version=gix_pack::data::Version::V2) at gix-pack/src/index/write/mod.rs:119
#25 0x0000002aab398a4e in gix_pack::Bundle::inner_write<&std::path::Path> (directory=..., progress=..., data_file=...,
    pack_entries_iter=..., should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>,
    pack_version=gix_pack::data::Version::V2) at gix-pack/src/bundle/write/mod.rs:287
#26 0x0000002aab399a20 in gix_pack::Bundle::write_to_directory<alloc::boxed::Box<gix_odb::Cache<gix_odb::store_impls::dynamic::Handle<alloc::sync::Arc<gix_odb::Store, alloc::alloc::Global>>>, alloc::alloc::Global>> (pack=...,
    directory=..., progress=..., should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>,
    thin_pack_base_object_lookup=..., options=...) at gix-pack/src/bundle/write/mod.rs:144
#27 0x0000002aab1ea34e in gix::remote::connection::fetch::Prepare<alloc::boxed::Box<(dyn gix_transport::client::blocking_io::traits::Transport + core::marker::Send), alloc::alloc::Global>>::receive_inner<alloc::boxed::Box<(dyn gix_transport::client::blocking_io::traits::Transport + core::marker::Send), alloc::alloc::Global>> (self=..., progress=...,
    should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>)
    at gix/src/remote/connection/fetch/receive_pack.rs:282
#28 0x0000002aab1ff04a in gix::clone::PrepareFetch::fetch_only_inner (self=0x3fffff42f8, progress=...,
    should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>) at gix/src/clone/fetch/mod.rs:199
#29 0x0000002aaad32568 in gix::clone::PrepareFetch::fetch_only<&mut prodash::progress::utils::DoOrDiscard<prodash::tree::Item>> (self=0x3fffff42f8, progress=0x3fffffa610, should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>)
    at gix/src/clone/fetch/mod.rs:75
#30 0x0000002aaad325ca in gix::clone::PrepareFetch::fetch_then_checkout<&mut prodash::progress::utils::DoOrDiscard<prodash::tree::Item>> (self=0x3fffff42f8, progress=0x3fffffa610,
    should_interrupt=0x2aac3b6e10 <gix::interrupt::IS_INTERRUPTED>) at gix/src/clone/fetch/mod.rs:231
#31 0x0000002aaae9269a in gitoxide_core::repository::clone::function::clone<prodash::progress::utils::DoOrDiscard<prodash::tree::Item>, std::ffi::os_str::OsString, std::path::PathBuf, &mut dyn std::io::Write, &mut dyn std::io::Write> (
    url=..., directory=..., overrides=..., progress=..., out=..., err=...) at gitoxide-core/src/repository/clone.rs:78
#32 0x0000002aaae36de6 in gitoxide::plumbing::main::main::{closure#13} (
    progress=<error reading variable: Cannot access memory at address 0x1>, out=..., err=...)
    at src/plumbing/main.rs:433
#33 0x0000002aaac43c3a in gitoxide::shared::pretty::prepare_and_run::{closure#0}<(), core::ops::range::RangeInclusive<u8>, gitoxide::plumbing::main::main::{closure_env#13}> () at src/shared.rs:170
#34 0x0000002aaac5e59c in gix_trace::enabled::Span::into_scope<core::result::Result<(), anyhow::Error>, gitoxide::shared::pretty::prepare_and_run::{closure_env#0}<(), core::ops::range::RangeInclusive<u8>, gitoxide::plumbing::main::main::{closure_env#13}>> (self=<error reading variable: Cannot access memory at address 0x0>,
    f=<error reading variable: Cannot access memory at address 0x31>) at gix-trace/src/lib.rs:43
#35 0x0000002aaac23680 in gitoxide::shared::pretty::prepare_and_run<(), core::ops::range::RangeInclusive<u8>, gitoxide::plumbing::main::main::{closure_env#13}> (name=..., trace=true, verbose=true, progress=false, progress_keep_open=false,
    range=..., run=...) at src/shared.rs:169
#36 0x0000002aaac0c38e in gitoxide::plumbing::main::main () at src/plumbing/main.rs:426
#37 0x0000002aaac0966c in gix::main () at src/gix.rs:5
(gdb)
```

This shows that the problem happens in `zlib-ng` and more specifically in `libz-ng-sys-1.1.15/src/zlib-ng/arch/riscv/adler32_rvv.c`, which through the `zlib-ng` repository resolves through the git submodule [to this file](https://github.com/zlib-ng/zlib-ng/blob/develop/arch/riscv/adler32_rvv.c#L15).

#### Likely explanation

This looks a lot like a case of https://github.com/zlib-ng/zlib-ng/issues/1705. My kernel is old enough to trigger that bug:

```text
ubuntu@riscv:~$ uname -a
Linux riscv 5.10.113-scw1 #1 SMP PREEMPT Fri Jul 12 15:31:22 UTC 2024 riscv64 riscv64 riscv64 GNU/Linux
```

Furthermore, that issue mentions the problem happening in starship, and I wonder if starship might even be triggering the crash in the same way through one of its `gix-*` dependencies.

Perhaps this issue is a victim of its own success and should be closed, to become (and be referenced by) a new comment on https://github.com/zlib-ng/zlib-ng/issues/1705. But I figured I'd start by opening this, in case more is known or can be discerned here, in case it might somehow help with #148, and in case there's anything to be done about it anywhere else (such as gitoxide).

#### Simplified reproduction

To check if the problem is occurring, and to check that it happens in `ein` as well as `gix`, the `gix status` and `ein t h` commands can be used:

```text
ubuntu@riscv:~/repos/gitoxide (main =)$ cargo run --bin=gix -- status
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.44s
     Running `target/debug/gix status`
Illegal instruction (core dumped)
ubuntu@riscv:~/repos/gitoxide (main =)[132]$ cargo run --bin=gix -r -- status
    Finished `release` profile [optimized] target(s) in 1.39s
     Running `target/release/gix status`
Illegal instruction (core dumped)
ubuntu@riscv:~/repos/gitoxide (main =)[132]$ cargo run --bin=ein -- t h
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.37s
     Running `target/debug/ein t h`
Illegal instruction (core dumped)
ubuntu@riscv:~/repos/gitoxide (main =)[132]$ cargo run --bin=ein -r -- t h
    Finished `release` profile [optimized] target(s) in 1.37s
     Running `target/release/ein t h`
Illegal instruction (core dumped)
```

EliahKagan:
This adds an `rvv-off` feature that prevents RVV vector instructions from being emitted in riscv64 builds even when the build machine is detected to support RVV. As discussed in https://github.com/rust-lang/libz-sys/issues/200#issuecomment-2283132037 and surrounding comments, this might be considered to fix #200.

I don't regard this to be in especially good shape, mostly because it only affects builds that use `cmake`. I did not include any changes to `cc.rs` to try to get it to work without `cmake`.

To make it easier to test this, I have also added recognition of an `RVV_OFF` environment variable with the same effect, whenever present with any value. Current use of this feature would often likely be to work around https://github.com/zlib-ng/zlib-ng/issues/1705 as described in #200, but this might also make it undesirable to use it *as a feature*, since it would involve modifying multiple crates in a transitive dependency chain. Using an environment variable is less reliable--at least currently implemented, rerunning `cargo` commands with a different value of the environment variable would not automatically rebuild anything--but more usable for this anticipated case. It also helped me in testing.

I am not sure about the names of either the feature or the environment variable. For the feature, I suspect it should have "experimental" in its name and not just its description, since `zlib-ng-no-cmake-experimental-community-maintained` does and this is at least as experimental as that (and also exacerbate that feature's experimental nature, but not actually having an effect when that feature is used).

I have not added anything in the test suite or to any CI scripts to test this. There seems to be a significant amount of functionality not under test in these ways, so maybe that is okay, but I am not sure.

*Edit:* Since opening this, I have noticed that the feature vs. environment variable considerations I was referring to have already been more precisely, clearly, and succinctly presented in https://github.com/rust-lang/libz-sys/pull/206#issuecomment-2294758685. (However, since the functionality here and in #206 differ significantly in how they would likely be used, I don't know if a decision in either PR about this would predict a decision in the other.)

---

To test this manually, I used `gitoxide`, since that is the software where I observed #200 as documented there. First I produced and populated a `libz-ng-sys` directory along the `libz-sys` working tree by running the `cargo-zng` script after temporarily modifying it in this way:

```diff
diff --git a/cargo-zng b/cargo-zng
index 4dd8ee3..ca072e3 100755
--- a/cargo-zng
+++ b/cargo-zng
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eu
-tempdir="$(mktemp -d)"
-trap 'rm -rf "$tempdir"' 0 INT
+tempdir=../libz-ng-sys; mkdir "$tempdir"  # tempdir="$(mktemp -d)"
+# trap 'rm -rf "$tempdir"' 0 INT
 cargo package -l --allow-dirty |
     tr '\\' '/' |
     grep -vxF -e Cargo.toml.orig -e .cargo_vcs_info.json |
@@ -10,5 +10,5 @@ cargo package -l --allow-dirty |
 cp Cargo-zng.toml "$tempdir/Cargo.toml"
 cp -a systest "$tempdir/systest"
 mv "$tempdir/systest/Cargo-zng.toml" "$tempdir/systest/Cargo.toml"
-cd "$tempdir"
-cargo "$@"
+# cd "$tempdir"
+# cargo "$@"
```

Then, in the `gitoxide` working tree, which was in the same parent directory as `libz-sys` and `libz-ng-sys`, I verified that `gix status` and a `gix clone` command both failed as described in #200 without patching in the changes here, as well as when they were patched in but nothing was set to enable them.

Then I patched them in and also set `RVV_OFF` to `1`. (I chose this value because it is intuitive. As noted above, the value is currently ignored.)

```sh
RVV_OFF=1 cargo build --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"'
```

[That fixed the crash, as hoped.](https://gist.github.com/EliahKagan/8177d0724ae8431bb108d9de59d98f98#file-exerpt-txt)

That shows `gix status`. A test with `gix clone`, as well as a number of tests to verify that without the change the problem occurs, i.e. to validate this manual testing procedure itself, appear in [this other longer and fairly disorganized transcript](https://gist.github.com/EliahKagan/8177d0724ae8431bb108d9de59d98f98#file-full-txt). (When I ran `vim` there, I did not use it to modify any files.)

### DIFF
--- a/Cargo-zng.toml
+++ b/Cargo-zng.toml
@@ -39,9 +39,12 @@ libc = "0.2.43"
 cmake = "0.1.50"
 
 [features]
-# This feature builds riscv64 binaries portable to machines without RVV, even if the build machine
-# is riscv64 and detected to support RVV. This is useful for building a more portable binary, or on
-# systems where auto-detection is incorrect (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
+# Experimental: This feature only affects zlib-ng, is only relevant to riscv64
+# targets, and currently only has an effect when `cmake` is used. It builds
+# binaries portable to machines without RVV, even if the build machine is
+# riscv64 and detected to support RVV. This is useful for building a more
+# portable binary, or on systems where auto-detection is incorrect
+# (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
 rvv-off = []
 
 [lints.rust]
--- a/Cargo-zng.toml
+++ b/Cargo-zng.toml
@@ -38,15 +38,5 @@ libc = "0.2.43"
 [build-dependencies]
 cmake = "0.1.50"
 
-[features]
-# Experimental: This feature only affects zlib-ng, is only relevant to riscv64
-# targets, and currently only has an effect when `cmake` is used. It builds
-# binaries portable to machines without RVV, even if the build machine is
-# riscv64 and detected to support RVV. This is useful for building a more
-# portable binary, or on systems where auto-detection is incorrect (as in
-# https://github.com/zlib-ng/zlib-ng/issues/1705). Building with the `RVV_OFF`
-# environment variable set (to any value) is equivalent to this feature.
-rvv-off = []
-
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(zng)'] }
--- a/Cargo-zng.toml
+++ b/Cargo-zng.toml
@@ -38,5 +38,11 @@ libc = "0.2.43"
 [build-dependencies]
 cmake = "0.1.50"
 
+[features]
+# This feature builds riscv64 binaries portable to machines without RVV, even if the build machine
+# is riscv64 and detected to support RVV. This is useful for building a more portable binary, or on
+# systems where auto-detection is incorrect (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
+rvv-off = []
+
 [lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(zng)'] }
--- a/Cargo-zng.toml
+++ b/Cargo-zng.toml
@@ -43,8 +43,9 @@ cmake = "0.1.50"
 # targets, and currently only has an effect when `cmake` is used. It builds
 # binaries portable to machines without RVV, even if the build machine is
 # riscv64 and detected to support RVV. This is useful for building a more
-# portable binary, or on systems where auto-detection is incorrect
-# (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
+# portable binary, or on systems where auto-detection is incorrect (as in
+# https://github.com/zlib-ng/zlib-ng/issues/1705). Building with the `RVV_OFF`
+# environment variable set (to any value) is equivalent to this feature.
 rvv-off = []
 
 [lints.rust]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -93,8 +93,9 @@ stock-zlib = []
 # targets, and currently only has an effect when `cmake` is used. It builds
 # binaries portable to machines without RVV, even if the build machine is
 # riscv64 and detected to support RVV. This is useful for building a more
-# portable binary, or on systems where auto-detection is incorrect
-# (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
+# portable binary, or on systems where auto-detection is incorrect (as in
+# https://github.com/zlib-ng/zlib-ng/issues/1705). Building with the `RVV_OFF`
+# environment variable set (to any value) is equivalent to this feature.
 rvv-off = []
 # Deprecated: the assembly routines are outdated, and either reduce performance
 # or cause segfaults.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,9 +89,10 @@ zlib-ng = ["libc", "cmake"]
 # or be completely removed in the future
 zlib-ng-no-cmake-experimental-community-maintained = ["libc"]
 stock-zlib = []
-# This feature only affects zlib-ng and is only relevant to riscv64 targets. It
-# builds binaries portable to machines without RVV, even if the build machine
-# is riscv64 and detected to support RVV. This is useful for building a more
+# Experimental: This feature only affects zlib-ng, is only relevant to riscv64
+# targets, and currently only has an effect when `cmake` is used. It builds
+# binaries portable to machines without RVV, even if the build machine is
+# riscv64 and detected to support RVV. This is useful for building a more
 # portable binary, or on systems where auto-detection is incorrect
 # (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
 rvv-off = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,12 @@ zlib-ng = ["libc", "cmake"]
 # or be completely removed in the future
 zlib-ng-no-cmake-experimental-community-maintained = ["libc"]
 stock-zlib = []
+# This feature only affects zlib-ng and is only relevant to riscv64 targets. It
+# builds binaries portable to machines without RVV, even if the build machine
+# is riscv64 and detected to support RVV. This is useful for building a more
+# portable binary, or on systems where auto-detection is incorrect
+# (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
+rvv-off = []
 # Deprecated: the assembly routines are outdated, and either reduce performance
 # or cause segfaults.
 asm = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,14 +89,6 @@ zlib-ng = ["libc", "cmake"]
 # or be completely removed in the future
 zlib-ng-no-cmake-experimental-community-maintained = ["libc"]
 stock-zlib = []
-# Experimental: This feature only affects zlib-ng, is only relevant to riscv64
-# targets, and currently only has an effect when `cmake` is used. It builds
-# binaries portable to machines without RVV, even if the build machine is
-# riscv64 and detected to support RVV. This is useful for building a more
-# portable binary, or on systems where auto-detection is incorrect (as in
-# https://github.com/zlib-ng/zlib-ng/issues/1705). Building with the `RVV_OFF`
-# environment variable set (to any value) is equivalent to this feature.
-rvv-off = []
 # Deprecated: the assembly routines are outdated, and either reduce performance
 # or cause segfaults.
 asm = []
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::ffi::OsStr;
 
 pub fn build_zlib_ng(target: &str, compat: bool) {
     let mut cmake = cmake::Config::new("src/zlib-ng");
@@ -14,8 +15,25 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             .define("WITH_DFLTCC_INFLATE", "1")
             .cflag("-DDFLTCC_LEVEL_MASK=0x7e");
     }
-    if target.contains("riscv64") && env::var_os("RVV_OFF").is_some() {
-        cmake.define("WITH_RVV", "OFF");
+    if target.contains("riscv") {
+        // Check if we should pass on an explicit boolean value of the WITH_RVV build option.
+        // See: https://github.com/zlib-ng/zlib-ng?tab=readme-ov-file#advanced-build-options
+        match env::var_os("RISCV_WITH_RVV")
+            .map(OsStr::to_str)
+            .map(str::trim)
+            .map(str::to_uppercase)
+            .map(Into::into)
+        {
+            Some("OFF" | "NO" | "FALSE" | "0") => {
+                // Force RVV off. This turns off RVV entirely, as well as the runtime check for it.
+                cmake.define("WITH_RVV", "OFF");
+            }
+            Some("ON" | "YES" | "TRUE" | "1") => {
+                // Try to use RVV, but still don't do so if a runtime check finds it unavailable.
+                // This has the same effect as omitting WITH_RVV, unless it has already been set.
+                cmake.define("WITH_RVV", "ON");
+            }
+        }
     }
     if target == "i686-pc-windows-msvc" {
         cmake.define("CMAKE_GENERATOR_PLATFORM", "Win32");
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -14,7 +14,7 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             .define("WITH_DFLTCC_INFLATE", "1")
             .cflag("-DDFLTCC_LEVEL_MASK=0x7e");
     }
-    if cfg!(feature = "rvv-off") && target.contains("riscv64") {
+    if target.contains("riscv64") && (cfg!(feature = "rvv-off") || env::var_os("RVV_OFF").is_some()) {
         cmake.define("WITH_RVV", "OFF");
     }
     if target == "i686-pc-windows-msvc" {
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -14,7 +14,7 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             .define("WITH_DFLTCC_INFLATE", "1")
             .cflag("-DDFLTCC_LEVEL_MASK=0x7e");
     }
-    if target.contains("riscv64") && (cfg!(feature = "rvv-off") || env::var_os("RVV_OFF").is_some()) {
+    if target.contains("riscv64") && env::var_os("RVV_OFF").is_some() {
         cmake.define("WITH_RVV", "OFF");
     }
     if target == "i686-pc-windows-msvc" {
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -26,6 +26,9 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
         {
             Some("OFF" | "NO" | "FALSE" | "0") => {
                 // Force RVV off. This turns off RVV entirely, as well as the runtime check for it.
+                // This is not usually necessary, but can be useful for building binaries portable
+                // to systems that do not support RVV but where auto-detection fails to identify
+                // this (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
                 cmake.define("WITH_RVV", "OFF");
             }
             Some("ON" | "YES" | "TRUE" | "1") => {
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -1,5 +1,4 @@
 use std::env;
-use std::ffi::OsStr;
 
 pub fn build_zlib_ng(target: &str, compat: bool) {
     let mut cmake = cmake::Config::new("src/zlib-ng");
@@ -18,23 +17,21 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
     if target.contains("riscv") {
         // Check if we should pass on an explicit boolean value of the WITH_RVV build option.
         // See: https://github.com/zlib-ng/zlib-ng?tab=readme-ov-file#advanced-build-options
-        match env::var_os("RISCV_WITH_RVV")
-            .map(OsStr::to_str)
-            .map(str::trim)
-            .map(str::to_uppercase)
-            .map(Into::into)
-        {
-            Some("OFF" | "NO" | "FALSE" | "0") => {
-                // Force RVV off. This turns off RVV entirely, as well as the runtime check for it.
-                // This is not usually necessary, but can be useful for building binaries portable
-                // to systems that do not support RVV but where auto-detection fails to identify
-                // this (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
-                cmake.define("WITH_RVV", "OFF");
-            }
-            Some("ON" | "YES" | "TRUE" | "1") => {
-                // Try to use RVV, but still don't do so if a runtime check finds it unavailable.
-                // This has the same effect as omitting WITH_RVV, unless it has already been set.
-                cmake.define("WITH_RVV", "ON");
+        if let Ok(value) = env::var("RISCV_WITH_RVV") {
+            match value.trim().to_uppercase().as_str() {
+                "OFF" | "NO" | "FALSE" | "0" => {
+                    // Force RVV off. This turns off RVV entirely, as well as the runtime check for it.
+                    // This is not usually necessary, but can be useful for building binaries portable
+                    // to systems that do not support RVV but where auto-detection fails to identify
+                    // this (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
+                    cmake.define("WITH_RVV", "OFF");
+                }
+                "ON" | "YES" | "TRUE" | "1" => {
+                    // Try to use RVV, but still don't do so if a runtime check finds it unavailable.
+                    // This has the same effect as omitting WITH_RVV, unless it has already been set.
+                    cmake.define("WITH_RVV", "ON");
+                }
+                _ => {}
             }
         }
     }
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -14,6 +14,9 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             .define("WITH_DFLTCC_INFLATE", "1")
             .cflag("-DDFLTCC_LEVEL_MASK=0x7e");
     }
+    if cfg!(feature = "rvv-off") && target.contains("riscv64") {
+        cmake.define("WITH_RVV", "OFF");
+    }
     if target == "i686-pc-windows-msvc" {
         cmake.define("CMAKE_GENERATOR_PLATFORM", "Win32");
     }


Cargo-zng.toml
```diff
@@ -38,5 +38,14 @@ libc = "0.2.43"
 [build-dependencies]
 cmake = "0.1.50"
 
+[features]
+# Experimental: This feature only affects zlib-ng, is only relevant to riscv64
+# targets, and currently only has an effect when `cmake` is used. It builds
+# binaries portable to machines without RVV, even if the build machine is
+# riscv64 and detected to support RVV. This is useful for building a more
+# portable binary, or on systems where auto-detection is incorrect
+# (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
```
Byron, line 47: I think here one would also have to advertise the `RVV_OFF` environment variable.

Cargo.toml
```diff
@@ -89,6 +89,13 @@ zlib-ng = ["libc", "cmake"]
 # or be completely removed in the future
 zlib-ng-no-cmake-experimental-community-maintained = ["libc"]
 stock-zlib = []
+# Experimental: This feature only affects zlib-ng, is only relevant to riscv64
+# targets, and currently only has an effect when `cmake` is used. It builds
+# binaries portable to machines without RVV, even if the build machine is
+# riscv64 and detected to support RVV. This is useful for building a more
+# portable binary, or on systems where auto-detection is incorrect
+# (as in https://github.com/zlib-ng/zlib-ng/issues/1705).
```
Byron, line 97: I think here one would also have to advertise the `RVV_OFF` environment variable.

Byron (changes requested):
Thanks a lot for tackling this!

What I like about this solution is that there is a documented cargo feature which (soon) would also advertise a corresponding environment variable. This resolves my biggest criticism towards using them - the difficulty to discover them outside of reading the code of build-scripts.

The only question left would be to what extend it has to be experimental. Is it the name that could change, or the way it's implemented, or is it the lack of `CC` support?
Under which circumstances would it be possible to remove the experimental status?
My answer to this would be "Once `cc.rs` also has support for it". As a matter of fact, if `flate2` could use `CC` instead of `cmake` in the `zlib-ng` backend, that would definitely make it more accessible and `gitoxide` would be easier to compile as well.

I'd also be looking forward to hearing what @jongiddy thinks about this.

Cargo-zng.toml
EliahKagan: I've added [d0d8002](https://github.com/rust-lang/libz-sys/pull/218/commits/d0d8002bedff3898ec6de6a269c3afb61a49fb97), which expands the comments to also document the `RVV_OFF` environment variable.

Cargo.toml
EliahKagan: I've made the same change in both places, per https://github.com/rust-lang/libz-sys/pull/218#discussion_r1749197667.

EliahKagan:
> What I like about this solution is that there is a documented cargo feature which (soon) would also advertise a corresponding environment variable. This resolves my biggest criticism towards using them - the difficulty to discover them outside of reading the code of build-scripts.

Would that be applicable to #206 as well then--could having both a feature and an environment variable there make sense?

If so, should the environment variables be named with a similar naming convention? I assume that, if done, this would entail keeping the current name there and expanding this one. Relatedly, maybe that would shed light on whether the value should be examined.

Byron (approved):
Yes, there it's a sole environment variable without documentation in `lib.rs` (on crate level at least for now). Here we started with something very specific to make it more suitable as a cargo feature, with the environment variable as another way to enable the feature to make it easier to enable for complex builds.

Admittedly, the only reason I am unsure about this is that I wonder what the original authors would want, or in lieu of this, what is most common with `*-sys` crates. To my mind, I'd not use environment variables at all unless they are strongly requested, and if that's done, they'd need proper documentation.

If this PR should be more like #206 instead, the environment variable could be named less specifically, the value could matter, and it could be documented on crate level instead. It seems to be more 'common' at least.

zng/cmake.rs
```diff
@@ -14,6 +14,9 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             .define("WITH_DFLTCC_INFLATE", "1")
             .cflag("-DDFLTCC_LEVEL_MASK=0x7e");
     }
+    if target.contains("riscv64") && (cfg!(feature = "rvv-off") || env::var_os("RVV_OFF").is_some()) {
```
jongiddy, line 17: Generally I prefer "positive" features over "negative" features. By this I mean that the feature sounds like it adds something rather than taking something away.

For example, you could name the environment variable `WITH_RVV` and allowing the values `ON` and `OFF`. This would be familiar to someone who knows the `zlib-ng` build. It would allow forcing the use of RVV if there was ever a case where it was disabled.

You could name the feature `with-rvv`. However this would require the default to compile without RVV. This has the advantage of working in more places, at the cost of behaving differently to the default behavior of `zlib-ng`.

jongiddy (commented):
I'm OK with the current PR. My one comment is mostly for consideration rather than requiring action.

I do suggest that a specific value is required for the environment variable, say `RVV_OFF=1`. An empty string `RVV_OFF=` should be the same as being unset. Any other values trigger an error. This allows flexibility if other variants are required.

Byron:
Thanks a lot for sharing your expertise, @jongiddy!

I'd be happy to go with all of the suggestions, while staying conservative on how the feature is handled so behaviour isn't changed. @EliahKagan are these changes you could do? And if present, please share your questions or suggestions as well.

EliahKagan:
It seems to me that there is no particularly good design available here, and that it might be best either to wait to see what progress looks like on https://github.com/zlib-ng/zlib-ng/issues/1705 (see https://github.com/zlib-ng/zlib-ng/pull/1770), or to generalize this into a mechanism for passing arbitrary `-D` options to `cmake`.

I think the problems here both for a feature (and its name) and for an environment variable (and its name) are that:

- Any change from the current default behavior--other than fixing [the bug that motivated this](https://github.com/zlib-ng/zlib-ng/issues/1705), which should be [done upstream](https://github.com/zlib-ng/zlib-ng/pull/1770), and which may not suffice for binaries that are to be redistributed--is actually taking build functionality away. The default behavior is to attempt to auto-detect if RVV instructions should be emitted. This behavior is, conceptually, a feature. Auto-detection usually works, but fails on some RISC-V systems without RVV on some kernels. Any way of replacing auto-detection with a pre-set choice is taking away that feature.
- Making the current behavior a feature, so RVV is forced off unless the feature is enabled, would make many builds generate slower code, such that users would likely assume there had simply been a performance regression and not investigate. Making it a default feature would still have this effect for builds where default features are disabled in order to avoid getting new default features, though perhaps that practice is rare. Making the current behavior a feature would also introduce skew between the `cmake` and non-`cmake` cases in a way that is very unintuitive--one would lose build functionality by switching to `cmake`.
- In another relevant sense, neither having RVV *nor omitting RVV* is "negative," because if RVV instructions are not emitted, it is not as though the relevant functionality is omitted: instead, it is implemented with other instructions (that are usually slower). This should in theory make it possible to name the feature something that doesn't sound negative. But because the alternative instructions can be anything otherwise supported by the assumed CPU features--it is not as though we would be using XYZ instructions instead of RVV instructions--it may be hard to name an RVV-omitting feature precisely without making it sound negative. `omit-rvv-for-compatibility` sounds more *active*, but it doesn't sound less negative.

So I don't know what to do. I agree with the points raised by @jongiddy, but I don't think there is any easy way to address them properly. We can change which behavior a feature represents, and even separately from that we can rename it so it seems positive rather than negative, but I fear the situation would be inelegant, confusing, and misleading in almost the same way as in the PR now.

Not adding a feature, and implementing functionality where an environment variable--or separate environment variables--can be used to pass arbitrary `-D` options to `cmake`, seems to me like it could be a way out. But that itself has three interrelated problems:

- It is a more general solution to a problem that is not known, in the setting of this crate, to have other cases where it is needed (though it could be convenient in other cases for experimentation).
- It feels like it should be implemented in the `cmake` crate instead, as something dependent crates like this one can use.
- If implemented, it should allow multiple `-D` options to be passed. Then either they are passed in the same environment variable and we are introducing our own parsing rules (even if those rules are just "split on whitespace and don't recognize quoting"); or they are passed in separate environment variables (e.g. `LIBZ_SYS_CMAKE_DEFINE_*`), in which case we have to decide whether to respect [case-folding](https://github.com/rust-lang/rust/blob/d7522d872601c5243899a813728a05cde1e5a8e2/library/std/src/sys/pal/windows/process.rs#L62-L64) for matching environment variable names on Windows.

I'd be happy to implement the more general functionality, but I'm checking first because I don't know if that is wanted.

Byron:
Thanks a lot for sharing, it feels like this conversation leads to a solution soon.

> What if we call the feature `skip-rvv-detection` and the environment variable `SKIP_RVV_DETECTION`? Or `skip-rvv-autodetection` and `SKIP_RVV_AUTODETECTION`?

To me that sounds preferable over `RVV_OFF`, and I assume that `WITH_RVV=OFF` as passed to cmake actually disables auto-detection only.

> Is that a good argument against continuing just to check for the existence of the environment variable? Actually I think it is not:

My intuition here was the opposite, primarily because I found it hard to decide what should override what. The solution seemed to be that if `SKIP_RVV_AUTODETECTION` is set to `OFF` or a similarly falsy value, it seems intuitive that it would override the cargo feature. The reason for this is that the feature can be controlled in the manifest, but the person building the software for distribution should have the final word on it.

So the logic would be something like `if (cargo_feature_set && !env_var_falsy) || env_var_truthy { skip-RVV }`. It should *probably* fail hard if the value of the environment variable isn't understood.

jongiddy:
All good points here. In order to make something available, may I suggest that we support the environment variable `USE_RISCV_RVV`. Adding the architecture makes it clear that this option is only relevant for RISC-V and disambiguates in case another architecture decides to use the acronym RVV.

My suggestions for allowed values are:
(unset), (empty), `auto`: auto-selection of RVV
`off`, `0`: force RVV disabled
`on`, `1`: force RVV enabled
All other values produce an error.

The `on` option is useful for people who know they will only run on RVV-enabled processors and want to be certain that they get the performance of RVV. If they happen to build on a non-RVV processor they would prefer the build to fail.

We can leave out the feature for now, hence skipping the question of priority.

EliahKagan:
This sounds good to me. I will look into whether there is a reasonable way to force RVV on or, if not, to force RVV instructions to be emitted.

Those two things are not quite the same. One problem, which I now recognize, is that I have not been properly distinguishing between [checks that take place when zlib-ng is built](https://github.com/zlib-ng/zlib-ng/blob/72c50edd26631c8709e91b5d4b58cf6df3beb521/CMakeLists.txt#L822-L843) and [checks that take place at runtime](https://github.com/zlib-ng/zlib-ng/blob/develop/arch/riscv/riscv_features.c) unless the build has prevented them from being done. The currently broken detection is in the latter (or at least that), and a fix based on [this workaround](https://github.com/zlib-ng/zlib-ng/issues/1705#issuecomment-2102075289) is being developed [in this pull request](https://github.com/zlib-ng/zlib-ng/pull/1770/files) which is [waiting on a separate change](https://github.com/zlib-ng/zlib-ng/pull/1770#issuecomment-2322904336) but largely done.

Maybe this was already clear based on the information in #200 and https://github.com/zlib-ng/zlib-ng/issues/1705 but, if not, I'm worried that a lot of the discussion here may be based on my misleading descriptions. I am furthermore not at all sure that I have had them straight in my own thinking: why have I so confidently claimed that a build with RVV forced off would be needed for redistribution, even after a *runtime* check is fixed? I am sorry about the confusion.

In any case, my plan now is to go forward [with that approach](https://github.com/rust-lang/libz-sys/pull/218#issuecomment-2342834201) unless it turns out not to be feasible or to involve excessive complexity, in which I'll go with something along previously discussed lines except still only with an environment variable.

EliahKagan:
It appears the only documented way to affect RVV in zlib-ng through build configuration is the `WITH_RVV` variable, defined in the [advanced build options](https://github.com/zlib-ng/zlib-ng?tab=readme-ov-file#advanced-build-options) table.

It is `cmake`-only: nothing appears for it in the `configure` column. This does not make it infeasible to achieve the same result without `cmake`, but since that goes beyond what the upstream project itself supports, I'm continuing not to attempt it as part of this PR.

The default is `ON`, and RVV instructions are emitted. Changing it to `OFF` avoids emitting RVV instructions (and thus also does not perform auto-detection). Forcing RVV instructions to be used even on systems that are detected as not supporting them does not appear to be an upstream feature, and looks like it would require editing the source code itself, beyond just configuration files. For this reason, I am departing from https://github.com/rust-lang/libz-sys/pull/218#issuecomment-2342834201, but still doing some parts of it:

- I am omitting the `rvv_off` feature. One of its benefits was the accompanying comments, where it was defined in the manifest files. The important information from those comments is reworded for clarity and appears in comments in `cmake.rs`.

- I am replacing the `RVV_OFF` environment variable whose value was not examined with a `RISCV_WITH_RVV` environment variable. Setting it to `ON`, `OFF`, or other Boolean synonyms, causes `WITH_RVV` to be defined with the value of `ON` or `OFF`. Otherwise it is not defined, which is equivalent to defining it with `ON` assuming nothing is in place to adjust its default value. Typically, one will either refrain from using this, or use the value `OFF`.

I am just ignoring empty or otherwise blank, as well as other values that don't easily parse to common truthy or falsy strings, treating any such values as if the environment variable had not been defined. This can be changed if desired.

I've tested that setting `RISCV_WITH_RVV` to `OFF` in the environment, with these changes, has the same effect as setting `RVV_OFF` to any value had achieved, before these changes. That is, with these commands, I verified that `RISC_WITH_RVV` can be used to suppress RVV and allow the gitoxide `gix status` and `gix clone` to work without SIGILL or any other problems:

```sh
RISCV_WITH_RVV=OFF cargo build --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"'
RISCV_WITH_RVV=OFF cargo run --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"' --bin=gix -- status
RISCV_WITH_RVV=OFF cargo run --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"' --bin=gix -- clone git@github.com:Byron/gitoxide.git
```

Byron (approved):
Thanks a lot, particularly for the additional description and with detailed reasoning.

When looking at [the docs](https://docs.rs/libz-sys/latest/libz_sys/) of this crate one sees that there is nothing hand-written. Hence, there is a strong precedent to also *not* require `RISCV_WITH_RVV` to be documented as part of the (nonexistent) docs.

If one feels inspired, I'd definitely value any kind of crate-level documentation that informs about the various backend (CC, Cmake) and the environment variables they may be influenced by.

EliahKagan:
One small correction to something I reported above. I mentioned also testing with:

```sh
RISCV_WITH_RVV=OFF cargo nextest run --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"' --all --no-fail-fast
GIX_TEST_IGNORE_ARCHIVES=1 RISCV_WITH_RVV=OFF cargo nextest run --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"' --all --no-fail-fast
```

But this is effectively the same as testing with just the first command:

```sh
RISCV_WITH_RVV=OFF cargo nextest run --config 'patch.crates-io.libz-sys.path="../libz-sys"' --config 'patch.crates-io.libz-ng-sys.path="../libz-ng-sys"' --all --no-fail-fast
```

That's because I didn't clean the directories extracted from generated archives in the first test run, so `GIX_TEST_IGNORE_ARCHIVES=1` in the second test run had no practical effect. Really this was just rerunning the tests.

Cleaning and running with `GIX_TEST_IGNORE_ARCHIVES=1` does [produce](https://gist.github.com/EliahKagan/0270e133dba98b62a113b5725d72cca4) one failure on the RISC-V test machine, but it is not related to the changes here, nor even to the upgrade to 2.2.2 in #219, because it [also happens](https://gist.github.com/EliahKagan/0270e133dba98b62a113b5725d72cca4#file-run-2-max-pure-txt) with `max-pure`, which I believe does not use crates from this project.

(The failing test is `gix-ref-tests::refs packed::iter::performance`, but I don't think my test machine is slow enough that the failure would be considered insignificant. I'll report it in `gitoxide` after a bit of further investigation.)

PR #221, opened 2024-10-02
ohos: Link shared lib on OpenHarmony

jschwe:
The OpenHarmony SDK / sysroot contains libz.so and similar to android we can assume that libz will always be available, 
since the [OpenHarmony zlib documentation](https://docs.openharmony.cn/pages/v5.0/en/application-dev/reference/native-lib/zlib.md) exists and explicitly mentions linking against `libz.so`.

OpenHarmony is supported as a tier 2 target in rustc as [*-linux-ohos](https://doc.rust-lang.org/rustc/platform-support/openharmony.html)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -23,9 +23,9 @@ fn main() {
     }
 
     // All android compilers should come with libz by default, so let's just use
-    // the one already there. Likewise, Haiku always ships with libz, so we can
-    // link to it even when cross-compiling.
-    if target.contains("android") || target.contains("haiku") {
+    // the one already there. Likewise, Haiku and OpenHarmony always ship with libz,
+    // so we can link to it even when cross-compiling.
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -60,7 +60,10 @@ fn main() {
                 }
             }
             Err(e) => {
-                println!("cargo:warning=Could not find zlib include paths via pkg-config: {}", e)
+                println!(
```
Byron, line 63: Could you undo this formatting change as it's unrelated?
*CI currently doesn't validate the format, and it seems `cargo fmt` skips `build.rs` entirely.*

Byron (changes requested):
Thanks a lot for the fix!

Once the requested change is made, I will merge.

Thank  you.

build.rs
jschwe: Reverted! FYI, this change was made by running `cargo fmt`.

Issue #230, opened 2024-12-05
zlib-ng-no-cmake-experimental-community-maintained breaks OSX builds


brainstorm:
Reproducer: https://github.com/wcampbell0x2a/backhand/issues/651

Byron:
I can reproduce it here as well:

```
> cargo build --no-default-features --features "zlib-ng-no-cmake-experimental-community-maintained"
   Compiling libz-sys v1.1.20 (/Users/byron/dev/github.com/rust-lang/libz-sys)
The following warnings were emitted during compilation:

warning: libz-sys@1.1.20: In file included from src/zlib-ng/arch/arm/adler32_neon.c:9:
warning: libz-sys@1.1.20: src/zlib-ng/arch/arm/neon_intrins.h:38:28: error: expected identifier or '('
warning: libz-sys@1.1.20:    38 | static inline uint16x8x4_t vld1q_u16_x4(uint16_t const *a) {
warning: libz-sys@1.1.20:       |                            ^
warning: libz-sys@1.1.20: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/16/include/arm_neon.h:10523:28: note: expanded from macro 'vld1q_u16_x4'
warning: libz-sys@1.1.20:  10523 | #define vld1q_u16_x4(__p0) __extension__ ({ \
warning: libz-sys@1.1.20:        |                            ^
warning: libz-sys@1.1.20: In file included from src/zlib-ng/arch/arm/adler32_neon.c:9:
warning: libz-sys@1.1.20: src/zlib-ng/arch/arm/neon_intrins.h:47:28: error: expected identifier or '('
warning: libz-sys@1.1.20:    47 | static inline uint8x16x4_t vld1q_u8_x4(uint8_t const *a) {
warning: libz-sys@1.1.20:       |                            ^
warning: libz-sys@1.1.20: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/16/include/arm_neon.h:10466:27: note: expanded from macro 'vld1q_u8_x4'
warning: libz-sys@1.1.20:  10466 | #define vld1q_u8_x4(__p0) __extension__ ({ \
warning: libz-sys@1.1.20:        |                           ^
warning: libz-sys@1.1.20: In file included from src/zlib-ng/arch/arm/adler32_neon.c:9:
[...]

PR #231, opened 2024-12-06
Disable neon for apple targets

Jake-Shadle:
Workaround for #230.

Don't really have time to look at this more in depth, so this just disables neon intrinsics for apple targets leaving the rest of them untouched. There's probably just a missing flag or something, but someone who actually has the will to compile for aarch64-apple-darwin can maybe figure that out.

### DIFF
--- a/zng/cc.rs
+++ b/zng/cc.rs
@@ -149,7 +149,7 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
             "crc32_braid_c",
             "crc32_fold_c",
             "slide_hash_c",
-        ]
+        ],
     );
 
     if compat {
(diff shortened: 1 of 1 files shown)


Jake-Shadle:
It would probably be good to add CI checking of `aarch64-apple-darwin`, apple targets are such a pain the ass I never bothered checking AFAIR, I'm fairly sure it has literally never worked, (ie I get the same errors on an outdated clang 13 machine).

Byron (approved):
Thanks a lot for your help, it's much appreciated!

A local test showed that it does indeed compile now.

Also, from what I can tell, it already is compiled on all platforms using a matrix, but it's allowed to fail when the `CC` feature is enabled. Even though warnings are emitted, thus far I admit to not have noticed them.

Jake-Shadle:
Not `aarch64-apple-darwin`, https://github.com/rust-lang/libz-sys/blob/223f5650c75fb4c0fec5e00fc47f2fd6b1a95cc2/.github/workflows/ci.yml#L79C9-L80C32

PR #235, opened 2025-02-10
Fix unnecessary search path warning on macOS

madsmtm:
The linker on macOS warns on unused search paths, and concretely, when compiling `zlib-ng` via this crate, it warns with:
```
ld: search path 'target/debug/build/libz-sys-3eb8b95c9c44bf93/out/lib64' not found
```

This warning has previously been hidden away, but `rustc` wants to surface them by default [at some point](https://github.com/rust-lang/rust/issues/136096).

So let's avoid the warning by setting `CMAKE_INSTALL_LIBDIR` explicitly, and then only passing a single search path to the linker.

Can be tested on macOS with:
```console
$ RUSTFLAGS="-Wlinker-messages" cargo +nightly test --no-default-features --features zlib-ng
```

### DIFF
--- a/zng/cmake.rs
+++ b/zng/cmake.rs
@@ -39,18 +39,18 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
         cmake.define("CMAKE_GENERATOR_PLATFORM", "Win32");
     }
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot! I validated he warning, and that it doesn't occur with this PR.

PR #237, opened 2025-03-02
bump `zlib-ng` to version `2.2.4`

folkertdev:
turns out zlib-ng had another release, with mostly bug fixes https://github.com/zlib-ng/zlib-ng/releases/tag/2.2.4.

Not urgent at all but we may as well stay up to date.


Byron (approved):
That seems like a good idea, thank you!

PR #244, opened 2025-07-02
fix: Remove forcing static builds when cross-compiling

polarathene:
This constraint is no longer necessary, due to improvements since it's original introduction in Feb 2016.

The zlib library will be linked dynamically (provided no explicit opt-in to static link) or fallback to building from source.

---

More details with discussion/history is covered here: https://github.com/rust-lang/libz-sys/pull/206#issuecomment-3026759631

**Notably:**
- A project should not link zlib both dynamically (_via another crate linking system deps that bring in zlib_) and statically (_vendored via `libz-sys`_): https://github.com/rust-lang/libz-sys/issues/201
- Building a target as a native arch or cross-compiled arch should not cause linking between dynamic vs static to occur when zlib is discoverable to link dynamically.

This PR ensures building for the target is consistent across correctly configured build environments.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -75,17 +75,12 @@ fn main() {
 
     // Situations where we build unconditionally.
     //
-    // MSVC basically never has it preinstalled, MinGW picks up a bunch of weird
-    // paths we don't like, `want_static` may force us, and cross compiling almost
-    // never has a prebuilt version.
-    //
-    // Apple platforms have libz.1.dylib, and it's usually available even when
-    // cross compiling (via fat binary or in the target's Xcode SDK)
-    let cross_compiling = target != host;
+    // - MSVC basically never has zlib preinstalled
+    // - MinGW picks up a bunch of weird paths we don't like
+    // - Explicit opt-in via `want_static`
     if target.contains("msvc")
         || target.contains("pc-windows-gnu")
         || want_static
-        || (cross_compiling && !target.contains("-apple-"))
     {
         return build_zlib(&mut cfg, &target);
     }
@@ -101,6 +96,7 @@ fn main() {
         return;
     }
 
+    // For convenience fallback to building zlib if attempting to link zlib failed
     build_zlib(&mut cfg, &target)
 }
 


polarathene:
I have another PR to contribute as a follow-up, but have realized it'll be less effective than I thought it'd be as `rustc` / Cargo behave a bit unexpectedly.

This PR that was merged will most likely affect those cross-compiling to musl from glibc if they did not use the `static` feature or equivalent ENV. I think that is ok as it's not uncommon to run into linking issues when building crates like that, but some crates do seem to detect musl target and treat it as build from source.

Either way, without a common ENV supported across crates, I think the concern is tricky to satisfy everyone 😓

Byron:
Thanks for the heads-up!

> I think that is ok as it's not uncommon to run into linking issues when building crates like that, but some crates do seem to detect musl target and treat it as build from source.

Do you have a prominent example for that? Maybe a big part of the downstream disruption could be averted if this crate would do something similar, based on what an established crate is already doing?

Thanks again.

polarathene:
> Do you have a prominent example for that?

I noticed the [`pcre2-sys` crate](https://github.com/BurntSushi/rust-pcre2/issues/7#issuecomment-1673045760) has done this, with the PR linked providing an opt-out via ENV for those that want to prevent that (_those packaging for musl distributions where dynamic linking is desired_).

My only concern with that approach is the more sys crates involved, the larger a growing list of ENV to opt-out would be for those affected. It would be similar for the opposite scenario of those wanting to always enforce static linking though, so the broader audience to support for the convenience of reduced ENV requirements is cross compiling to musl targets implies static by convention in Rust?

Another crate [`libpng-sys`](https://github.com/kornelski/rust-libpng-sys/blob/main/build.rs#L12-L17) detects static and always builds instead of linking to a prebuilt `.a` static archive because in 2019 a user reported a build failure as the static library they had to link on CentOS 7 wasn't compiled with `-fPIC`, however [there was a misunderstanding as this is not required by Rust to link](https://github.com/kornelski/rust-libpng-sys/issues/1#issuecomment-3041050469) for an executable the relocation model just needed to change from default PIC to static.

---

It's possible unify static preference via an explicit `RUSTFLAGS='-C target-feature=+crt-static'` (_the opposite `-crt-static` however is not, and it is not possible to detect an implicit `+crt-static` either with targets like musl, thus not entirely consistent_). Which is an approach I implemented in my PR prior to discovering it wasn't detected implicitly with musl targets.

Rust also has `-C prefer-dynamic=no` as a code-gen option I think, but I'd need to look into that and not entirely sure how useful it'd be in a `build.rs` script if it can't be detected 🤔 

As soon as I find some extra time I'll open a PR within the next day with what I had done so far and figure out a way forward from there 👍

PR #251, opened 2025-11-10
update to `zlib-ng` version `2.2.5`

folkertdev:


folkertdev:
As a heads-up, zlib-ng did a bunch of modifications to their cmake files and compile setup, so users may see some new linker errors etc. when building from source. We're encountering a couple, at least: https://github.com/trifectatechfoundation/zlib-rs/actions/runs/19329002658.

I suspect this will be rare in practice, but it's something to keep an eye on.

Byron:
Thanks for sharing!

It's interesting that the CI here didn't notice this, and I wonder if that's a shortcoming.
Thus far I haven't heard of issues, but if they multiple we can always yank the update.

Issue #158, opened 2023-09-21
Dynamically link zlib-ng


hack3ric:
Linux distros prefer shared libraries over vendor it whenever possible, and zlib-ng is (at least) [available on Arch Linux](https://archlinux.org/packages/extra/x86_64/zlib-ng/). I am wondering if we can support dynamic linking in libz-ng-sys?

I would like to help and create a PR, but we may need to discuss how to expose this feature first, if it is acceptable.

Byron:
Even though I don't know enough yet to help with figuring out implementation details, I can say that any implementation of this feature would need to be backwards compatible, i.e. non-breaking.

With this said, such a feature will also help to one day make it possible to let people opt-in to using system libraries, static or dynamic, via `flate2` as well.

NobodyXu:
Maybe it could use a CARGO environment for this instead of a feature?

I've seen other *-sys crates (e.g. openssl) doing this, with a vendored feature to override environment settings.

chenrui333:
might be just do `ZLIB_NO_VENDOR` like `openssl` and `libgit2` crates?

folkertdev:
For zlib-rs we want to test against locally-built versions of zlib a lot, so I hacked up a script that takes the `libz-sys` source, but transforms all of the extern functions into a function that dynamically loads from a dylib.

https://github.com/memorysafety/zlib-rs/blob/main/dynamic-libz-sys/src/generated.rs

e.g.

```rust
extern "C" {
    fn deflateReset(strm: z_streamp) -> c_int;
}
```
becomes
```rust
pub unsafe fn deflateReset(strm: z_streamp) -> c_int {
    type Func = unsafe extern "C" fn(strm: z_streamp) -> c_int;
    let f: libloading::Symbol<Func> = dynamic_library()
        .get(if_zng!({ concat!("zng_", "deflateReset") }, "deflateReset").as_bytes())
        .unwrap();
    f(strm)
}
```

The reason this approach is useful for us is that we can load two versions of zlib into the same binary without name conflicts. There is a performance penalty I guess but we only need this for debugging anyway so that's not an issue. Maybe this is useful for someone while this crate does not implement picking a custom dylib to link.

PR #252, opened 2025-11-14
fix: Support opt-out of static linking via `LIBZ_SYS_STATIC=0`

polarathene:
This change refactors the `want_static` logic to additionally support `LIBZ_SYS_STATIC=0` as an [explicit opt-out to the `static` feature when `libz-sys` is a transitive dependency](https://github.com/rust-lang/libz-sys/issues/201#issuecomment-2291519891) and the build environment prefers to dynamic link `libz` instead.

I've inlined some rough commentary should it help grokking the helper method to maintainers (_and any future contributors_), with clear intent of supporting opt-out from the crate's `static` feature. This could probably do with some revision, along with [user visible documentation](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-2325610571) of ENV precedence over the crate feature.

---

## Future work

A follow-up PR may want to further refactor the following with [consideration of `LIBZ_SYS_STATIC` explicitly forcing a linkage preference](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-2294758685) as opposed to convenience fallback/assumptions (_the commentary was contributed some time ago and lacks further context/sources_):

https://github.com/rust-lang/libz-sys/blob/80c597a07f5e5f0dff0e98c03d48a77845de9f6e/build.rs#L25-L31

https://github.com/rust-lang/libz-sys/blob/80c597a07f5e5f0dff0e98c03d48a77845de9f6e/build.rs#L39-L42

https://github.com/rust-lang/libz-sys/blob/80c597a07f5e5f0dff0e98c03d48a77845de9f6e/build.rs#L76-L86

https://github.com/rust-lang/libz-sys/blob/80c597a07f5e5f0dff0e98c03d48a77845de9f6e/build.rs#L88-L100

Likewise a follow-up PR may want to consider [dynamic link support for `zlib-ng`](https://github.com/rust-lang/libz-sys/issues/158):

https://github.com/rust-lang/libz-sys/blob/80c597a07f5e5f0dff0e98c03d48a77845de9f6e/build.rs#L16-L23

The zlib logic already enforces a build if `want_static == true`, otherwise runs a test to check if a system library is available to link, and if not falls back to building from source.

## References

For anyone engaging on the topic/support, these might be of interest:
- [Historical context](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-3026759631)
- [Removal of enforced static linking for cross-compilation](https://github.com/rust-lang/libz-sys/pull/244)
- [Cross-compilation example with pkg-config](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-3026759898)
- [Conventions used by other sys crates](https://github.com/rust-lang/libz-sys/pull/206#issuecomment-2325455490)
  - `<CRATE>_NO_VENDOR=1` (git, openssl)
  - `<CRATE>_USE_PKG_CONFIG=1` (ssh, zstd, sqlite)
  - `<CRATE>_STATIC=0` (pcre2, libz)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -238,11 +238,11 @@ fn zlib_installed(cfg: &mut cc::Build) -> bool {
     false
 }
 
-// The environment variable `LIBZ_SYS_STATIC` is first checked for a value of `0` (false) or `1` (true),
-// before considering the `static` feature when no explicit ENV value was detected.
-// When `libz-sys` is a transitive dependency from a crate that forces static linking via the `static` feature,
-// this enables the build environment to revert that preference via `LIBZ_SYS_STATIC=0`.
-// The default is otherwise `false`.
+/// The environment variable `LIBZ_SYS_STATIC` is first checked for a value of `0` (false) or `1` (true),
+/// before considering the `static` feature when no explicit ENV value was detected.
+/// When `libz-sys` is a transitive dependency from a crate that forces static linking via the `static` feature,
+/// this enables the build environment to revert that preference via `LIBZ_SYS_STATIC=0`.
+/// The default is otherwise `false`.
 fn should_link_static() -> bool {
   let has_static_env: Option<&'static str> = option_env!("LIBZ_SYS_STATIC");
   let has_static_cfg = cfg!(feature = "static");


Byron:
Thanks so much, this is fantastic work!

Let's keep working in this direction, to become comfortable to make the changes this needs to evolve.

Something I'd be looking forward to is to document the current behaviour in such a way that the documentation can be kept in sync with future changes. Maybe adding docs can happen, in parts, once the respective parts of the build script are refactored, so it's an incremental process as well.
Maybe… the build-script could be mostly self-documenting as well, with detailed doc strings, and the `lib.rs` docs only documents the big picture so people can control their builds without having to study the source.

All just ideas, let's see what happens next :).

PR #257, opened 2026-02-17
upgrade to zlib-ng 2.3.3

folkertdev:
https://github.com/zlib-ng/zlib-ng/releases/tag/2.3.3

This fixes https://github.com/zlib-ng/zlib-ng/pull/2102, which was first reported on zlib-rs.

### DIFF
--- a/zng/cc.rs
+++ b/zng/cc.rs
@@ -163,7 +163,7 @@ pub fn build_zlib_ng(target: &str, compat: bool) {
         fs::create_dir_all(&build).unwrap();
         build.push("gzread.c");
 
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot!

I validated that commit 127310 is indeed associated with the tag 2.3.3:

```
1273109 2026-02-03 22:10 +0100 Hans Kristian Rosbach              o [HEAD] {origin/2.3.x} {origin/stable} <2.3.3> 2.3.3 Bugfix release
```

Also I assume that `gzread_mangle.h.in` is the new name and it wouldn't compile without this change.

PR #259, opened 2026-02-25
Bump zlib submodule to 1.3.2

dalvescb:
The current version of `zlib` is vulnerable to https://nvd.nist.gov/vuln/detail/CVE-2026-27171

Requesting a bump to 1.3.2, I've seen no issues building with it

EDIT: Corrected CVE to CVE-2026-27171,

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -119,6 +119,9 @@ fn build_zlib(cfg: &mut cc::Build, target: &str) {
 
     if !cfg!(feature = "libc") || target.starts_with("wasm32") {
         cfg.define("Z_SOLO", None);
+        // zlib 1.3.2 uses `NULL` directly in compress.c/uncompr.c, but the
+        // Z_SOLO config path doesn't pull in headers that always define it.
+        cfg.define("NULL", Some("0"));
     } else {
         cfg.file("src/zlib/gzclose.c")
             .file("src/zlib/gzlib.c")
@@ -126,8 +129,11 @@ fn build_zlib(cfg: &mut cc::Build, target: &str) {
             .file("src/zlib/gzwrite.c");
     }
 
+    // zlib's zconf.h gates standard headers on `STDC` (not `__STDC__`), and
+    // 1.3.2 uses `NULL` directly in compress.c/uncompr.c.
+    cfg.define("STDC", None);
+
     if !target.contains("windows") {
-        cfg.define("STDC", None);
         cfg.define("_LARGEFILE64_SOURCE", None);
         cfg.define("_POSIX_SOURCE", None);
         cfg.flag("-fvisibility=hidden");


Byron:
The strange thing is that the checks log for the 1.3.2 commit shows that the CI passed on some of the platforms that are failing here! This might hint at something we can do in the build script maybe.

Byron:
Thanks a lot for chiming in @tbeu!

Indeed, I think this PR makes no sense on the grounds of fixing a vulnerability. However, it might have an impact improving Zlib's downstream compatibility, let's see what comes out of it.

If they decline it, I think we need to go forward with this change to support zlib in this and future versions.

Meanwhile, I put it back into draft until it's clear what to do.

Byron:
Ok, I had it researched and it seems we should just merge this.

---

  CVE-2026-27171 / GHSA-h858-mf2m-8jf4 says zlib before 1.3.2 can hit unbounded CPU consumption in crc32_combine64 and crc32_combine_gen64 when given a negative length. GitHub marks it low severity and unreviewed, and upstream fixed it in commit ba829a4 with exp
  licit negative-length checks. On March 1, 2026, Mark Adler confirmed that commit is the fix for the CVE. Sources: https://github.com/advisories/GHSA-h858-mf2m-8jf4, https://github.com/madler/zlib/issues/904#issuecomment-3981149680,
  https://github.com/madler/zlib/commit/ba829a458576d1ff0f26fc7230c6de816d1f6a77

  For libz-sys specifically:

  - origin/main and release 1.1.24 still point the vendored stock-zlib submodule at 51b7f2a, which is zlib v1.3.1, so the bundled copy is in the affected version range.
  - libz-sys exposes crc32_combine:348, and upstream stock zlib implements that via crc32_combine64(... crc32_combine_gen64(len2)), so the exposed API does reach the vulnerable path.
  - But libz-sys does not always use the vendored copy. Its build script prefers a system zlib when available, and only builds bundled stock zlib in cases like Windows/MSVC, MinGW, explicit static builds, or fallback when no system zlib is found. See build.rs:16,
    build.rs:34, build.rs:74, and build.rs:83.

  So the precise answer is:

  - It does apply to libz-sys when libz-sys builds or ships its bundled stock zlib 1.3.1.
  - It does not apply when libz-sys links against a system zlib that is already patched to 1.3.2 or distro-backported equivalent.
  - It does not apply to zlib-ng mode.

  Practical nuance: this is a narrow issue. It requires calling the combine API with an invalid negative length, which upstream originally described as caller misuse, not a normal decompression path. So I would treat it as a real but low-severity issue, not a
  panic situation. Still, for the crate release and for scanner noise, updating the bundled zlib to 1.3.2 is the right move.

  That leaves my recommendation unchanged: merge PR 259 after correcting the PR description to reference CVE-2026-27171, not the earlier mistaken CVE.

PR #268, opened 2026-04-22
[AIX] Use zlibNX if available.

amy-kwan:
This change prevents libz-sys from building its own zlib by using zlibNX if available, which is an version of the zlib library that supports hardware accelerated data compression and decompression.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -30,19 +30,6 @@ fn main() {
         return;
     }
 
-    // On AIX, use zlibNX if available.
-    if target.contains("aix") {
-        let zlibNX_lib = "/usr/opt/zlibNX/lib";
-        let zlibNX_include = "/usr/opt/zlibNX/include";
-        if std::path::Path::new(zlibNX_lib).join("libz.a").exists() &&
-           std::path::Path::new(zlibNX_include).join("zlib.h").exists() {
-            println!("cargo:rustc-link-search=native={}", zlibNX_lib);
-            println!("cargo:rustc-link-lib=z");
-            println!("cargo:include={}", zlibNX_include);
-            return;
-        }
-    }
-
     let want_static = should_link_static();
     // Don't run pkg-config if we're linking statically (we'll build below) and
     // also don't run pkg-config on FreeBSD/DragonFly. That'll end up printing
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
Thanks a lot. Here is a review by Codex, based on my hunch, which ended up P1. Please do also take a look at P2, which may or may not make sense.

Thanks

----

- **P1: Honor explicit static mode before preferring zlibNX**  
  In [build.rs](/Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:33), the new AIX fast-path returns as soon as `/usr/opt/zlibNX` exists, before `should_link_static()` is evaluated. That changes existing behavior: callers using the `static` feature or `LIBZ_SYS_STATIC=1` can no longer force the bundled static build when zlibNX is present. This regresses a supported build mode.

- **P2: Validate zlibNX is actually linkable before skipping fallback**  
  In [build.rs](/Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:37), the AIX special case short-circuits based only on file existence. The current non-Windows system path only avoids the vendored build after `zlib_installed()` successfully compiles and links `src/smoke.c`. If `/usr/opt/zlibNX` exists but is unusable for the active compiler configuration, this change will fail later at link time instead of falling back to the bundled zlib.

amy-kwan:
> Thanks a lot. Here is a review by Codex, based on my hunch, which ended up P1. Please do also take a look at P2, which may or may not make sense.
> 
> Thanks
> 
> * **P1: Honor explicit static mode before preferring zlibNX**
>   In [build.rs](/Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:33), the new AIX fast-path returns as soon as `/usr/opt/zlibNX` exists, before `should_link_static()` is evaluated. That changes existing behavior: callers using the `static` feature or `LIBZ_SYS_STATIC=1` can no longer force the bundled static build when zlibNX is present. This regresses a supported build mode.
> * **P2: Validate zlibNX is actually linkable before skipping fallback**
>   In [build.rs](/Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:37), the AIX special case short-circuits based only on file existence. The current non-Windows system path only avoids the vendored build after `zlib_installed()` successfully compiles and links `src/smoke.c`. If `/usr/opt/zlibNX` exists but is unusable for the active compiler configuration, this change will fail later at link time instead of falling back to the bundled zlib.

Ah, those are good points. Thank you for the review, @Byron! Appreciate it!
I've attempted to address these comments in the newest commit.

Byron (commented):
Thanks a lot, looks good to me.

Codex came up with something else given only the new commit, but I can't understand what it means, it seems off.

With that said, can you please confirm that you validated that this works by hand on an aix system? Please also share interactions with want_static and maybe anything else you can throw its way.

Thank you.

----

- [P2] Check zlibNX before emitting pkg-config metadata — /Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:87-93
  On AIX systems where zlibNX is installed but `pkg-config` also finds another `zlib.pc` (for example under `/opt/freeware`), the probe above this block has already emitted `cargo:rustc-link-search`/`cargo:rustc-link-lib` lines before this zlibNX path returns. Those stale linker directives can put the other zlib search path ahead of `/usr/opt/zlibNX/lib`, so the build no longer reliably prefers zlibNX; the zlibNX check should happen before any pkg-config metadata is emitted while still remaining after the `want_static` early return.

Byron:
@amy-kwan Maybe you can also take a look at #275 and see if it relates to this PR at all.

PR #270, opened 2026-05-27
forward the `crc` target feature on aarch64

folkertdev:
When statically linking `zlib` (i.e. when building it from source), forward the `crc` target feature. `zlib` does not dynamically detect it, and apparently it is not part of the baseline feature set in C.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
         platform: [linux-musl]
 
   windows:
-    runs-on: windows-latest
+    runs-on: ${{ matrix.runner || 'windows-latest' }}
     # Windows technically doesn't need this, but if we don't block windows on it
     # some of the windows jobs could fill up the concurrent job queue before
     # one of the install-cross jobs has started, so this makes sure all
@@ -55,11 +55,14 @@ jobs:
           - x86_64-pc-windows-gnu
           - i686-pc-windows-msvc
           - i686-pc-windows-gnu
+          - aarch64-pc-windows-msvc
         include:
           - target: x86_64-pc-windows-gnu
             mingw_package: mingw-w64-x86_64-gcc
           - target: i686-pc-windows-gnu
             mingw_package: mingw-w64-i686-gcc
+          - target: aarch64-pc-windows-msvc
+            runner: windows-11-arm
 
   macos:
     runs-on: macos-latest
--- a/build.rs
+++ b/build.rs
@@ -137,6 +137,17 @@ fn build_zlib(cfg: &mut cc::Build, target: &str) {
         cfg.flag("-fvisibility=hidden");
     }
 
+    // Forward target features to the C compiler.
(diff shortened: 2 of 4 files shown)


Byron (changes requested):
Thanks a lot! This PR seems like something I should have merged before making a new release, but here we are 😅.

As now there is no rush, maybe you can take a look at this Codex-generated comment. It looks like CI wouldn't detect the issue. Maybe this is also something this PR could add, maybe even first, to reproduce the issue (please see the second block).

I am setting the PR to changes requested, just because I don't have other means of labeling it. All it needs is your feedback/investigation.

----

[P2] Guard CRC march flag for MSVC compilers — /Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:146-147
  When stock zlib is built for an AArch64 MSVC target with `-C target-feature=+crc`, this raw GCC/Clang `-march=armv8-a+crc` flag is passed to the C compiler. `cl.exe` does not accept that option, so `aarch64-pc-windows-msvc` builds that enable CRC will fail in the build script; gate this on the compiler family or use a supported/probed flag path.

----

No, this issue should not show up in the current CI configuration.

The review comment looks valid as a latent bug: MSVC targets always build bundled zlib via [build.rs](/Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:82), and the CRC path unconditionally passes GCC/Clang syntax at [build.rs](/Users/byron/dev/github.com/rust-lang/libz-sys/build.rs:146).

But CI does not cover the failing combination:

- Windows CI only tests `x86_64`/`i686` MSVC/GNU targets, not `aarch64-pc-windows-msvc`: [.github/workflows/ci.yml](/Users/byron/dev/github.com/rust-lang/libz-sys/.github/workflows/ci.yml:53)
- Linux CI tests AArch64, but only Linux targets: [.github/workflows/ci.yml](/Users/byron/dev/github.com/rust-lang/libz-sys/.github/workflows/ci.yml:138)
- I found no workflow/config setting `RUSTFLAGS=-Ctarget-feature=+crc` or similar.
- `rustc --print cfg --target aarch64-pc-windows-msvc` reports `target_feature="neon"` by default, not `crc`.

So CI would only catch this if it added an `aarch64-pc-windows-msvc` build with CRC enabled, for example via `RUSTFLAGS="-Ctarget-feature=+crc"`.

folkertdev:
Later commits 

- add a test for `--features=static`
- suppress a warning in `systest` on msvc (from what I understand it verifies the rust bindings against the C signatures)
- run the test suite on MSVC aarch64 windows

folkertdev:
ping @Byron

Byron (approved):
Sorry for the hiatus, that was just bad timing, and I'm now catching up on everything I have missed.

With that said, I took a look at every commit and am thankful for the added test coverage. With that I think everything is addressed.

Also, there now is an aarch64 run on Windows, which is great for peace of mind.

With that, I think we should be good to go, also assuming that you could validate locally that the CRC feature forwarding does give an observable speedup.

PR #273, opened 2026-07-24
Only define Z_SOLO on wasm32-unknown-unknown, not the wasi/emscripten targets

calvinrp:
`build_zlib()` defines `Z_SOLO` for every `wasm32*` target. `Z_SOLO` is
zlib's "no C library" mode: no default `zalloc`/`zfree` — so the one-shot
helpers `compress()`/`uncompress()` always return `Z_STREAM_ERROR` at
runtime — and no `gz*` API.

Only `wasm32-unknown-unknown` actually lacks a libc. The wasi targets
link wasi-libc (full malloc/free and fd-based I/O) and emscripten ships
musl, so there the blanket define just breaks working API: on
`wasm32-wasip1`,

```rust
let rc = libz_sys::compress(dst.as_mut_ptr(), &mut dst_len,
                            src.as_ptr(), src.len() as _);
assert_eq!(rc, libz_sys::Z_OK); // fails today: rc == -2 (Z_STREAM_ERROR)
```

compiles fine and fails at runtime on every call, while the identical
program works on native targets. flate2 doesn't notice because it always
supplies its own allocators through the streaming API; direct libz-sys
users on wasi hit it with no workaround inside libz-sys.

Minimal repro (any wasi runtime, e.g. wasmtime):

```
cargo new zsolo && cd zsolo && cargo add libz-sys@1.1.29
# main.rs: call libz_sys::compress() on any buffer and print the rc
cargo build --target wasm32-wasip1 && wasmtime target/wasm32-wasip1/debug/zsolo.wasm
# prints rc = -2 (Z_STREAM_ERROR); the same program prints 0 on native
```

This PR scopes the define to `wasm32-unknown-unknown`. wasi/emscripten
now take the standard source list, including the `gz*` files, which
compile cleanly against wasi-libc (verified on wasm32-wasip1: the crate
builds, and a one-shot compress/uncompress roundtrip passes where it
returned -2 before). `src/lib.rs` needs no change — it already declares
the one-shot functions unconditionally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,20 @@ jobs:
       # The threads ABI imports shared memory, which needs a custom host.
       - run: cargo test --target wasm32-wasip1-threads --test wasm --no-run
 
+  wasm-emscripten:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER: node
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          submodules: true
+      - uses: emscripten-core/setup-emsdk@4528d102f7230f0e7b276855c01ea1159be0e984 # v16
+        with:
+          version: 6.0.4
+      - run: rustup target add wasm32-unknown-emscripten
+      - run: cargo test --target wasm32-unknown-emscripten --test wasm --features static
+
   windows:
     runs-on: ${{ matrix.runner || 'windows-latest' }}
     # Windows technically doesn't need this, but if we don't block windows on it
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,23 @@ jobs:
       matrix:
         platform: [linux-musl]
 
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for contributing! This seems to be good, but it also doesn't look like we have any WASM tests here.

@jongiddy do you think it's feasible to try to add WASM test coverage? If not, or if you don't know, do you think it's safe to merge this PR as is? Thanks you. (PS: I didn't investigate this with an agent yet).

folkertdev:
In zlib-rs we test with custom cargo runners with a wide range of target architectures, mostly with qemu but we also include wasmtime for webassembly.

https://github.com/trifectatechfoundation/zlib-rs/blob/5a82af8a944d3b2d23e861c5438bac35bc40d16d/.github/workflows/checks.yaml#L22-L115

Byron:
Thanks for chiming in @folkertdev! 
I will give it a shot.

Byron:
This seems to work - I will have the other WASM runtimes added as well so we can test all environments that are affected by this change.

Byron:
@folkertdev Thanks again! There are now tests for a bunch of WASM runtimes, which validate the claims that were originally made here. Thus it appears like the PR is truly an improvements, and thanks to CI it should stay that way.

